PR #1, opened 2021-05-31
Update `ttt-reborn` into a working state

mzegar:
Spent some time updating this gamemode into a working state. This changelist is extremely rough, however, my goal with this PR was just to get the base code into a working state. You'll see things that probably should be renamed, mass comments of code, etc.

Quick rough crop showing the player actually loads.


saibotk:
Thank you for your effort :) 

Awesome to see that you got it to run. 
I'm made sure to assign both of the team that have worked on the code and I'm sure they will review it as soon as possible.

Thank you very much :)

code/Game.cs
```diff
@@ -1,11 +1,12 @@
 ﻿using System;
 using Sandbox;
 using System.Collections.Generic;
+using System.Linq;
 using System.Threading.Tasks;
 
 namespace TTTGamemode
 {
-    [ClassLibrary("ttt-reborn", Title = "Trouble in Terry's Town")]
+    [Library("ttt", Title = "Trouble in Terrorist Town")]
```
ZenBre4ker, line 9: Why do you change the name, we've chosen?

mzegar: Having a dash in the library name actually stops the project from launching completely, so trimmed off the `-reborn` part. Let me know if `tttreborn` maybe looks better.

The `.addon` folder has the title `Trouble in Terrorist Town` which is what users see, wanted to keep it consistent so changed it from `Trouble in Terry's Town` to `Trouble in Terrorist Town`
https://github.com/TTTReborn/ttt-reborn/blob/bc2819e1cde422a1f2b729a45e0572041efc56ea/.addon#L20

Let me know and I can update both copies to whatever the reborn team sees fit.

ZenBre4ker: Well as far as Im informed we want to go full-in with ttt reborn and Terry's Town to be more "modern" and give youtubers the possibility to not get demonetized for writing terrorist in their title.
So I guess you could change that already, but you can also wiat for more approval of the core developers.

TimGoll: Yep, what @ZenBre4ker says is correct. We want `Trouble in Terry's town`. It still has the abbrevation TTT, but the name is slightly different while also fitting to S&Box.

Alf21: Agreed. Additionally, I'd like to use "tttreborn" instead of just "ttt" to avoid issues with other TTT projects for S&Box as we can't claim the exact name ours imo (regarding BKU's agreement).

code/player/TTTPlayer.cs
```diff
@@ -5,26 +5,41 @@
...
         private DamageInfo _lastDamageInfo;
 
-        public Player()
+        public TTTPlayer()
         {
-            Inventory = new Inventory(this);
-            Animator = new StandardPlayerAnimator();
+	        Inventory = new Inventory(this);
 
-            Role = Role.None;
```
Alf21, line 24: Yea you renamed it to avoid issues with the `Role` enum (something I documented in the Trello as well), but I think `Role` would be more intuitive. Instead, we probably should move or rename the `Role` enum.

Alf21 (changes requested):

code/Game.cs
```diff
@@ -50,38 +58,36 @@ private void ChangeRound(Round round)
...
+                    
+                    int detectiveCount = (int) (All.Count * 0.125f);
+                    int traitorCount = (int) Math.Max(All.Count * 0.25f, 1f);
 
-                    #region Select Roles
-                    int detectiveCount = (int) ((float) Sandbox.Player.All.Count * 0.125f);
-                    int traitorCount = (int) Math.Max((float) Sandbox.Player.All.Count * 0.25f, 1f);
-
-                    List<Player> _players = Sandbox.Player.ConvertAll(p => (Player) p);
+                    var players = Client.All.ToList().ConvertAll(p => p.Pawn as TTTPlayer);
```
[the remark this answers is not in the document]
mzegar: `Player.All` refers to all of the player's character entities. I don't also like 

```csharp
var players = Client.All.ToList().ConvertAll(p => p.Pawn as TTTPlayer);
```
But, I think it's fine to leave as is for now until some helper function is written, or your team decides how they want to handle it.

code/player/Body.cs
```diff
@@ -9,7 +9,7 @@ namespace TTTGamemode
 {
     public class Body : ModelEntity
     {
-        public Player Player { get; set; }
```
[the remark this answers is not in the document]
mzegar: Used inside of `Player.cs`, `TickInspectBody` being one of them.

code/Game.cs
```diff
@@ -193,21 +197,19 @@ private async Task StartGameTimer()
 
         private void UpdateGameTimer()
         {
-            CheckMinimumPlayers();
+	        CheckMinimumPlayers();
```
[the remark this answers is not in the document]
mzegar: Seems aligned with the rest of the file, could be github?

code/Game.cs
```diff
@@ -50,38 +58,36 @@ private void ChangeRound(Round round)
...
                     Random random = new Random();
 
                     // SELECT DETECTIVES
-                    for (int i = 0; i < detectiveCount; i++)
+                    for (var i = 0; i < detectiveCount; i++)
                     {
-                        int randomID = random.Next(_players.Count);
-                        _players[randomID].PlayerRole = Player.Role.Detective;
+                        var randomId = random.Next(players.Count);
+                        players[randomId].CurrentRole = Role.Detective;
```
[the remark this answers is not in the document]
mzegar:
```
Yea you renamed it to avoid issues with the Role enum (something I documented in the Trello as well), but I think Role would be more intuitive. Instead, we probably should move or rename the Role enum.
```

Not sure if I understand exactly, but...

I've moved the `RoleType` enum inside of the `Player` class.

code/Game.cs
```diff
@@ -217,37 +219,31 @@ public override void PostLevelLoaded()
...
-
-                return;
-            }
-
-            base.PlayerJoined();
+	        base.ClientJoined(client);
+	        
+	        var player = new TTTPlayer();
+	        Karma.RegisterPlayer(player);
+	        client.Pawn = player;
```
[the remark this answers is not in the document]
mzegar: `Client` & `Pawn` are both interfaces. It seems like the go to in all FB gamemodes is just to set the attribute.

https://github.com/Facepunch/sbox-hidden/blob/de76e5b894953eeb37044ae733d54eace22cc88e/code/Game.cs#L153

Alf21 (approved):
Thanks for your work! We just need to fix the identation issues so far, as it seems you are using tabs directly. Working with GitHub, it seems to be the best using spaces imo, so most of all, I'm using a 4 spaces indentation.

mzegar:
> Thanks for your work! We just need to fix the identation issues so far, as it seems you are using tabs directly. Working with GitHub, it seems to be the best using spaces imo, so most of all, I'm using a 4 spaces indentation.

Apologies, will change it in the future.

PR #4, opened 2021-06-02
[WIP] Polymorphic based round system

mzegar:
Still some minor changes to make, so WIP... wanted to make sure the `Reborn` team is fine with the structure/change.

**Changes**
- Adds a very similar round system found within the [FP Hidden gamemode](https://github.com/Facepunch/sbox-hidden/tree/main/code/rounds)
- This seems to be a pretty reasonable way to handle rounds. All round logic is found inside an inheritable abstract class, instead of being handled inside of `Game.cs`. Also `BaseRound.cs` basically provides an interface for potentially creating custom rounds for the future.


code/Properties/launchSettings.json
```diff
@@ -1,9 +1,9 @@
...
-            "executablePath": "C:\\Program Files (x86)\\Steam\\steamapps\\common\\sbox\\sbox.exe",
-            "commandLineArgs": "+developer +sw -noassert -tools"
-        }
+  "profiles": {
+    "ttt-reborn": {
+      "commandName": "Executable",
+      "executablePath": "C:\\Program Files (x86)\\Steam\\steamapps\\common\\sbox\\sbox.exe",
+      "commandLineArgs": "\u002Bdeveloper \u002Bsw -noassert -tools"
     }
+  }
```
TimGoll, line 8: is there a reason you switched back to 2 space indentation here?

code/rounds/BaseRound.cs
```diff
@@ -0,0 +1,92 @@
+using Sandbox;
+using System;
+using System.Collections.Generic;
+using TTTReborn.Player;
```
TimGoll, line 4: we like to add an empty line between gamemode namespaces and our namespaces

code/rounds/BaseRound.cs
```diff
@@ -0,0 +1,92 @@
...
+            {
+                Players.Add( player );
+            }
+        }
+
+        public virtual void OnPlayerSpawn( TTTPlayer player ) { }
+
+        public virtual void OnPlayerKilled( TTTPlayer player ) { }
+
+        public virtual void OnPlayerLeave( TTTPlayer player )
```
TimGoll, line 63: please don't add spaces around brackets here. I see you are inconsistent with this in your own file, probably something defined by your editor. So just a heads up

code/rounds/InProgressRound.cs
```diff
@@ -0,0 +1,112 @@
...
+        
+        public override void OnPlayerLeave( TTTPlayer player )
+        {
+            base.OnPlayerLeave( player );
+
+            Spectators.Remove( player );
+
+            if ( IsRoundOver() )
+            {
+                _ = ChangeToPostRound();
```
TimGoll, line 42: I've seen this quite a few times and I don't understand the purpose of `_`. Could you explain this?

TimGoll (commented):
I've looked through your code and I really like the idea. I added a few comments, mainly about the code style, nothing major though. Thank you!

code/Properties/launchSettings.json
mzegar: No reason, IDE messing with me.

code/rounds/BaseRound.cs
mzegar: Yup, please feel free to call out any weird coding syntax. I'm currently used to automatically formatting, so I tend to currently leave weird syntax all over the place.

code/rounds/InProgressRound.cs
mzegar: It tells the compiler that you don't plan on using the return value for anything. Used in the FP code as well, so I'm assuming it is useful to do.

code/Properties/launchSettings.json
mzegar: Agh, my IDE for whatever reason is continuing to mess with this file. Anyways, should we just add this file to our `gitignore`? Each of us will have a unique `executablePath`.

Alf21: Yea guess it's the generation tool / csproj. Anyways, doesn't matter. I agree, please add this file to the gitignore, so it is still in this repo but won't update in the upcoming time :)

code/ttt-reborn.csproj
```diff
@@ -3,8 +3,8 @@
 	<PropertyGroup>
 		<TargetFramework>net5.0</TargetFramework>
 		<GenerateDocumentationFile>true</GenerateDocumentationFile>
-		<AssemblyName>tttreborn</AssemblyName>
-		<PackageId>tttreborn</PackageId>
+		<AssemblyName>ttt-reborn</AssemblyName>
+		<PackageId>ttt-reborn</PackageId>
 		<LangVersion>preview</LangVersion>
 		<NoWarn>1701;1702;1591;</NoWarn>
 	</PropertyGroup>
```
mzegar, line 10: Pretty sure this is `GenerateProjects.bat` doing this, not related to my IDE.

code/Properties/launchSettings.json
Histalek: that is not how .gitignore works ^^ once a file is tracked by git, the .gitignore does nothing.

That file would need to be explicitly removed **and** put into the .gitignore

PR #8, opened 2021-06-03
Meta: Added codestyle to .editorconfig & format files

Histalek:
These additions are directly taken from dotnet runtime repository [1]. I did not go through all of them by hand so if you see something that you want to be handled differently ( e.g. naming styles, spaces ) please tell me.

All .editorconfig statements can be found in the microsoft documentation's language rules [2] and formatting rules [3].

[1] https://github.com/dotnet/runtime
[2] https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/style-rules/language-rules
[3] https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/style-rules/formatting-rules

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -14,3 +14,138 @@ insert_final_newline = true
 [*.md]
 insert_final_newline = false
 trim_trailing_whitespace = false
+
+# C# files
+[*.cs]
+# New line preferences
+csharp_new_line_before_open_brace = all
+csharp_new_line_before_else = true
+csharp_new_line_before_catch = true
+csharp_new_line_before_finally = true
+csharp_new_line_before_members_in_object_initializers = true
+csharp_new_line_before_members_in_anonymous_types = true
+csharp_new_line_between_query_expression_clauses = true
+
+# Indentation preferences
+csharp_indent_block_contents = true
+csharp_indent_braces = false
+csharp_indent_case_contents = true
+csharp_indent_case_contents_when_block = true
+csharp_indent_switch_labels = true
+csharp_indent_labels = one_less_than_current
+
+# Types: use keywords instead of BCL types, and permit var only when the type is clear
+csharp_style_var_for_built_in_types = false:suggestion
+csharp_style_var_when_type_is_apparent = false:none
+csharp_style_var_elsewhere = false:suggestion
+dotnet_style_predefined_type_for_locals_parameters_members = true:suggestion
+dotnet_style_predefined_type_for_member_access = true:suggestion
+
+# name all constant fields using PascalCase
+dotnet_naming_rule.constant_fields_should_be_pascal_case.severity = suggestion
+dotnet_naming_rule.constant_fields_should_be_pascal_case.symbols  = constant_fields
+dotnet_naming_rule.constant_fields_should_be_pascal_case.style    = pascal_case_style
+dotnet_naming_symbols.constant_fields.applicable_kinds   = field
+dotnet_naming_symbols.constant_fields.required_modifiers = const
+dotnet_naming_style.pascal_case_style.capitalization = pascal_case
+
+# static fields should have s_ prefix
+dotnet_naming_rule.static_fields_should_have_prefix.severity = suggestion
+dotnet_naming_rule.static_fields_should_have_prefix.symbols  = static_fields
+dotnet_naming_rule.static_fields_should_have_prefix.style    = static_prefix_style
+dotnet_naming_symbols.static_fields.applicable_kinds   = field
+dotnet_naming_symbols.static_fields.required_modifiers = static
+dotnet_naming_symbols.static_fields.applicable_accessibilities = private, internal, private_protected
+dotnet_naming_style.static_prefix_style.required_prefix = s_
+dotnet_naming_style.static_prefix_style.capitalization = camel_case
+
+# internal and private fields should be _camelCase
+dotnet_naming_rule.camel_case_for_private_internal_fields.severity = suggestion
+dotnet_naming_rule.camel_case_for_private_internal_fields.symbols  = private_internal_fields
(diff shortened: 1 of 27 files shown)


.editorconfig
```diff
@@ -14,3 +14,138 @@ insert_final_newline = true
...
+csharp_style_throw_expression = true:suggestion
+csharp_style_conditional_delegate_call = true:suggestion
+
+# Other features
+csharp_style_prefer_index_operator = false:none
+csharp_style_prefer_range_operator = false:none
+csharp_style_pattern_local_over_anonymous_function = false:none
+
+# Space preferences
+csharp_space_after_cast = false
```
TimGoll, line 124: we previously used a space after cast. I don't know if we should continue doing so, but I wanted to mention it

TimGoll (approved):
looks good to me, I left one comment to discuss

.editorconfig
Histalek: image

looking at the visual representation a space definitely looks better. More feedback/reactions would be nice, but i'm inclined to change it

TimGoll: But I don't think it is that important because we decided to use `int y = x as int` anyways

Histalek: yesn't. `as` does not work for primitive types right? so this rule would still be important i think

TimGoll: Oh. This might be true, I don't know to be honest. We never used it for primary types

.editorconfig
```diff
@@ -14,3 +14,138 @@ insert_final_newline = true
...
+csharp_space_after_colon_in_inheritance_clause = true
+csharp_space_after_comma = true
+csharp_space_after_dot = false
+csharp_space_after_keywords_in_control_flow_statements = true
+csharp_space_after_semicolon_in_for_statement = true
+csharp_space_around_binary_operators = before_and_after
+csharp_space_around_declaration_statements = do_not_ignore
+csharp_space_before_colon_in_inheritance_clause = true
```
saibotk, line 132: Before? Doesnt that look a bit weird?

saibotk (changes requested):
Looks good otherwise

saibotk (approved):

.editorconfig
TimGoll: you probably noticed it yourself because you resolved it, but for everyone else, here's the reason: https://docs.microsoft.com/de-de/dotnet/fundamentals/code-analysis/style-rules/formatting-rules#csharp_space_before_colon_in_inheritance_clause

Histalek (commented):
just 2 things i've looked up and i'm not sure about what we want to use

.editorconfig
```diff
@@ -14,3 +14,138 @@ insert_final_newline = true
...
+csharp_space_between_empty_square_brackets = false
+csharp_space_between_method_call_empty_parameter_list_parentheses = false
+csharp_space_between_method_call_name_and_opening_parenthesis = false
+csharp_space_between_method_call_parameter_list_parentheses = false
+csharp_space_between_method_declaration_empty_parameter_list_parentheses = false
+csharp_space_between_method_declaration_name_and_open_parenthesis = false
+csharp_space_between_method_declaration_parameter_list_parentheses = false
+csharp_space_between_parentheses = false
+csharp_space_between_square_brackets = false
+
```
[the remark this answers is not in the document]
TimGoll: agreed. Even though it works without, especially more complex conditional statements quickly become hard to read

.editorconfig
```diff
@@ -14,3 +14,138 @@ insert_final_newline = true
...
+csharp_new_line_before_finally = true
+csharp_new_line_before_members_in_object_initializers = true
+csharp_new_line_before_members_in_anonymous_types = true
+csharp_new_line_between_query_expression_clauses = true
+
+# Indentation preferences
+csharp_indent_block_contents = true
+csharp_indent_braces = false
+csharp_indent_case_contents = true
+csharp_indent_case_contents_when_block = true
```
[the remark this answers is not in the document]
saibotk: I am in favor of keeping this enabled

.editorconfig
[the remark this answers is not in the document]
saibotk: I think thats fine

.editorconfig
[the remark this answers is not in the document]
TimGoll: I honestly have no opinion on this

.editorconfig
Histalek: another point for clarification, these are ordered lexicographically so a bit higher up there is the corresponding `csharp_after_colon_in_inheritance_clause`. Combined they conclude to this
```c#
class C : I
{

}
```

TimGoll: yes, they should both be true

TimGoll:
From my point of view it can be merged

Histalek:
i still want to keep it open a bit longer because only three of us have voiced their opinion on this

.editorconfig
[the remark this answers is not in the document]
Alf21: I'm okay with both, but if I have to say what I prefer, I'd use `never_if_unnecessary` as it makes things more complicated with unnecessary chars. It's clear that `&&` has a higher priority than `||`. With things that definitely need brackets, it could be annoying to read. But that should be a rare case anyways, so I don't care as much  ^^

.editorconfig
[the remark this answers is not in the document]
Alf21: I'm definitely against this as it mismatches the current coding style. The brackets are in the same layer as the `case`, you can't do anything else below, no matter whether you are using additional indentation or not. So no reason using the first version. I'd use it like every other logical condition. Anyways, if there are multiple `case`s in a switch-statement, I'd like to see an empty line between them.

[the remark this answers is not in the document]
mzegar: Agree with Alf on this one.

Alf21 (approved):

PR #9, opened 2021-06-03
UI: Added PlayerInfo, GameTimer and WeaponSelection HUD elements

Alf21:
image

The PlayerInfo is full functional, GameTimer and WeaponSelection are hardcoded.

## TODO
- [ ] Functional GameTimer (Round rework got merged)


code/ui/Crosshair.cs
```diff
@@ -99,6 +103,27 @@ public Panel SetupCrosshair(Properties crosshairProperties)
...
+
+        if (weapon == null)
+        {
+            this.Style.Display = DisplayMode.None;
+        }
+        else
+        {
+            this.Style.Display = DisplayMode.Flex;
+        }
+    }
```
mzegar, line 126: Would it be better to have a `empty/hands` weapon that ensures no crosshair is created instead of checking each tick?

Here's an example of how `DM98` creates crosshairs for each of their weapons.
https://github.com/Facepunch/dm98/blob/68d346bca7084b3b8cced71d6f75e38803cc9d35/code/weapons/DmWeapon.cs#L240

if you agree, we can create this `empty/hands` weapon in a separate PR also, since it isn't really in the scope of this.

code/ui/GameTimer.cs
```diff
@@ -0,0 +1,23 @@
...
+{
+    public class GameTimer : Panel
+    {
+        public Label TimeLabel { set; get; }
+
+        public GameTimer()
+        {
+            StyleSheet.Load("/ui/GameTimer.scss");
+
+            TimeLabel = Add.Label("00:00", "timelabel");
```
mzegar, line 14: Not sure if the team discussed, should we have some system of adding `TODO`? Since this needs to be hooked up later.

code/ui/PlayerInfo.cs
```diff
@@ -38,16 +37,72 @@ public override void Tick()
...
+            }
+
+            HealthBar.TextLabel.Text = $"{player.Health:n0}";
+
             var weapon = player.ActiveChild as Weapon;
 
             if (weapon != null)
             {
-                Weapon.Text = $"{weapon.AmmoClip}";
+                if (invisibleAmmo)
```
mzegar, line 77: Does `invisibleAmmo` mean that the current weapon has no ammo ? i.e melee weapon?

Would a name like `HasAmmo` be better? or something along the lines of that.

mzegar (commented):
Minor comments, the UI looks really awesome so far.

code/ui/Crosshair.cs
Alf21: Yea this just was planned as a "hotfix", a better implementation is to create / update the Crosshair on WeaponSwitch etc. :)

code/ui/GameTimer.cs
Alf21: https://trello.com/b/Q5xOoF4Y/ttt-reborn

code/ui/PlayerInfo.cs
Alf21: Yea currently a dirty workaround as well as long as there isn't any Inventory System :D But yea, you are right, can be directly connected to the Weapon (anyways, should just trigger on ammo change of a weapon).

code/ui/GameTimer.cs
mzegar: ah right, neat.

code/ui/Crosshair.cs
mzegar: Yup no worries, feel free to leave it as is.

code/ui/PlayerInfo.cs
mzegar: Marking as resolved, as long as we just fix it in the future.

code/ui/PlayerInfo.cs
```diff
@@ -38,16 +37,72 @@ public override void Tick()
...
+            RoleLabel.Text = $"{player.Role.ToString()}";
+        }
+    }
+
+    public class IndicatorsPanel : Panel
+    {
+        public BarPanel HealthBar { set; get; }
+        public BarPanel AmmoBar { set; get; }
+
+        public bool invisibleAmmo { set; get; }
```
TimGoll, line 49: ok, this might be due to me being new to C#, but why is `invisibleAmmo` lower case?

TimGoll (approved):

mzegar (approved):
Tim has one comment, otherwise LGTM.

PR #15, opened 2021-06-10
Role System: Adding dynamic polymorphic role system

Alf21:


code/player/Player.RPCs.cs
```diff
@@ -0,0 +1,19 @@
+using Sandbox;
+
+using TTTReborn.Roles;
+
+namespace TTTReborn.Player
+{
+    partial class TTTPlayer
+    {
+        /// <summary>
+        /// Called by the server, updates client's `Role`.
```
mzegar, line 10: Very nit picky...

"Must be called on the server, updates TTTPlayer's `Role`.

code/player/Player.cs
```diff
@@ -2,16 +2,30 @@
...
+                {
+                    SetRole(To.Single(this), role.Name);
+                }
+            }
+            get {
+                return role;
+            }
+        }
+
+        private BaseRole role = new NoneRole();
```
mzegar, line 28: Unless I'm misunderstanding, where do we keep track of the players role on the server?

mzegar (commented):
Quick question

TimGoll:
Just a general question. I personally prefer naming variables and classes with the "type" in the beginning. So instead of "NoneRole" it would be "RoleNone". What's the general opinion on this?

code/player/Player.RPCs.cs
Alf21: Hmm. Good idea regarding the missing separation between Client and Server methods.

code/player/Player.cs
Alf21: Currently, the role on the server is the role the client has. No client is aware of the role of other players, but I thought that's something for another PR as it gonna be complex (storing information the client is aware of, displaying other roles to the clients than they actually have, etc.).

Alf21:
> Just a general question. I personally prefer naming variables and classes with the "type" in the beginning. So instead of "NoneRole" it would be "RoleNone". What's the general opinion on this?

hmmm,... You don't say ArrayList ArrayListPlayer. Most of all, you're saying playerList, right? ^^

mzegar:
> > Just a general question. I personally prefer naming variables and classes with the "type" in the beginning. So instead of "NoneRole" it would be "RoleNone". What's the general opinion on this?
> 
> hmmm,... You don't say ArrayList ArrayListPlayer. Most of all, you're saying playerList, right? ^^

In the case of `Role`, I personally think `TraitorRole` rolls off the tongue a bit better.

Alf21:
> > > Just a general question. I personally prefer naming variables and classes with the "type" in the beginning. So instead of "NoneRole" it would be "RoleNone". What's the general opinion on this?
> 
> > 
> 
> > hmmm,... You don't say ArrayList ArrayListPlayer. Most of all, you're saying playerList, right? ^^
> 
> 
> 
> In the case of `Role`, I personally think `TraitorRole` rolls off the tongue a bit better. We should however typically avoid the type in the name where possible.

Then how about just "Traitor" and "Innocent"? ^^

mzegar (approved):
LGTM, I'm fine with just "Traitor" tbh.

PR #16, opened 2021-06-11
Fea: Basic inventory mock

mzegar:
Branch of https://github.com/TTTReborn/ttt-reborn/pull/14

Not sure if we had any other ideas to implement the inventory, but just wanted to stub some stuff in to make it playable. There is def a better way to improve this.


code/player/Inventory.cs
```diff
@@ -0,0 +1,92 @@
...
+
+        int GetNextIndex(int index, int maxIndex)
+        {
+            return NormalizeSlotIndex(index + 1, maxIndex);
+        }
+
+        public int NormalizeSlotIndex(int index, int maxIndex)
+        {
+            return index > maxIndex ? 0 : (index < 0 ? maxIndex : index);
+        }
```
mzegar, line 46: Can we set some of these `helper` methods to be private inside the class?

code/player/Inventory.cs
```diff
@@ -0,0 +1,92 @@
...
+
+        public int GetNextWeaponSlot(WeaponType weaponType = 0)
+        {
+            int listSize = Count();
```
mzegar, line 50: This name isn't very descriptive. 

`inventoryLength`? `inventorySize`? `inventoryCount`? `weapons`? `items`?

code/player/Inventory.cs
```diff
@@ -0,0 +1,92 @@
...
+            if (listSize < 1)
+            {
+                return -1;
+            }
+
+            int activeSlot = GetActiveSlot();
+
+            if (activeSlot != -1)
+            {
+                int nextIndex = GetNextIndex(activeSlot, listSize - 1);
```
mzegar, line 61: I'm pretty sure we could just make `GetNextIndex` to be specific to `Inventory`.

```csharp
        int GetNextInventoryIndex(int index)
        {
            return NormalizeSlotIndex(index + 1, Count() - 1);
        }
```

code/player/Inventory.cs
```diff
@@ -0,0 +1,92 @@
...
+                Weapon nextWeapon = List[nextIndex] as Weapon;
+
+                if (weaponType != 0)
```
mzegar, line 64: What does `0` represent? Can we make this the actual enum name?

code/player/Inventory.cs
```diff
@@ -0,0 +1,92 @@
...
+                        // if there is no weapon with same slot or no slot defined, return the next available weapon
+                        return nextIndex;
+                    }
+
+                    return -1;
+                }
+            }
+
+            // edge case, if List does not contain the active weapon
+            return 0;
```
mzegar, line 89: How does `-1` and `0` differ? Might be worth leaving some docs ontop of the function.

I'm assuming `0` means no weapon, and `-1` means error?

code/player/Player.cs
```diff
@@ -78,13 +82,14 @@ public override void OnKilled()
...
+
+            Inventory.DropActive();
             Inventory.DeleteContents();
         }
 
         public override void Simulate(Client client)
         {
-            SimulateActiveChild(client, ActiveChild);
-
+            // Input requested a weapon switch
```
mzegar, line 92: Move comment down right?

code/player/Player.cs
```diff
@@ -98,6 +103,26 @@ public override void Simulate(Client client)
...
+            if (Input.Pressed(InputButton.Drop) && ActiveChild != null && Inventory != null)
+            {
+                int weaponSlot = (int) (ActiveChild as Weapon).WeaponType;
+                Entity droppedEntity = Inventory.DropActive();
+
+                if (droppedEntity != null)
+                {
+                    if (droppedEntity.PhysicsGroup != null)
+                    {
+                        droppedEntity.PhysicsGroup.Velocity = Velocity + (EyeRot.Forward + EyeRot.Up) * 300;
```
mzegar, line 115: Let's save this `300` magic number as a static variable in this class. We can probably make a ttt server param for it as well.

code/ui/WeaponSelection.cs
```diff
@@ -1,37 +1,188 @@
...
+            if (player == null)
+            {
+                return;
+            }
+
+            // update weapon slots, check for removed weapons etc.
+            Inventory inventory = player.Inventory as Inventory;
+            int inventoryCount = inventory.Count();
+            List<Weapon> toAdd = new();
+            List<WeaponSlot> tmpSlots = new();
```
mzegar, line 42: Hmm could be more descriptive names, trying to come up with myself too.
```csharp
List<Weapon> newWeapons = new();
List<WeaponSlot> ...
```

code/ui/WeaponSelection.cs
```diff
@@ -1,37 +1,188 @@
...
+                foreach (WeaponSlot weaponSlot in weaponSlots.Values)
+                {
+                    weaponSlot.Delete();
+                }
+
+                weaponSlots.Clear();
+
+                inventory.List.Sort(delegate(Entity wep1, Entity wep2) {
+                    return (wep1 as Weapon).WeaponType.CompareTo((wep2 as Weapon).WeaponType);
+                });
```
mzegar, line 78: Does C# really require a delegate, can you not just provide a lambda function?

code/ui/WeaponSelection.cs
```diff
@@ -1,37 +1,188 @@
...
+                    // add in order
+                    weaponSlots.Add(weapon.Name, new WeaponSlot(this, weapon));
+                }
+
+                // update for dropped active weapon maybe
+                oldActiveWeapon = null;
+            }
+
+            // update current selection
+            Weapon activeWeapon = player.ActiveChild as Weapon;
```
mzegar, line 92: `// setup` doesn't provide us anymore info than the actual code itself. 

Another example, `activeWeapon` tells us exactly that the comment mentions.
```csharp
// update current selection
Weapon activeWeapon = player.ActiveChild as Weapon;
```

mzegar (commented):
I understand that it's still WIP, but some comments anyways. Looking good so far.

code/player/Inventory.cs
Alf21: Just `GetNextIndex`

code/player/Inventory.cs
Alf21: It's the size of `List` of that `Inventory` class :D

code/player/Inventory.cs
Alf21: There is no WeaponType == 0, that's why it's 0 ^^
I aligned the weapon types with the slot numbers to make it simple.

code/player/Player.cs
Alf21: no ^^

code/ui/WeaponSelection.cs
Alf21: Yea, but that's still a dirty way. Theoretically, we have to fetch which weapon was dropped (per RPCs?), then remove the corresponding WeaponSlot and same if a weapon was added (sort the DOM again, somehow ^^)

code/ui/WeaponSelection.cs
Alf21: Old habits ^^

code/player/Inventory.cs
```diff
@@ -0,0 +1,102 @@
...
+            if (entity is Weapon weapon)
+            {
+                Sound.FromWorld("dm.pickup_weapon", entity.Position);
+            }
+
+            bool added = base.Add(entity, makeActive);
+
+            List.Sort((Entity wep1, Entity wep2) => {
+                return (wep1 as Weapon).WeaponType.CompareTo((wep2 as Weapon).WeaponType);
+            });
```
mzegar, line 34: List.Sort((Entity wep1, Entity wep2) => (wep1 as Weapon).WeaponType.CompareTo((wep2 as Weapon).WeaponType));

mzegar, line 34: `List.Sort((Entity wep1, Entity wep2) => (wep1 as Weapon).WeaponType.CompareTo((wep2 as Weapon).WeaponType));`

Could also remove curly braces, but this works as well.

code/player/Inventory.cs
```diff
@@ -0,0 +1,102 @@
...
+
+            return added;
+        }
+
+        public bool IsCarryingType(Type t)
+        {
+            return List.Any(x => x.GetType() == t);
+        }
```
mzegar, line 42: Do we want to make this private?

code/player/Inventory.cs
```diff
@@ -0,0 +1,102 @@
...
+
+        /// <summary>
+        /// Returns the next weapon slot. If a `WeaponType` is given and the corresponding key was pressed, the index of a different weapon with
+        /// the same slot will be returned.
+        /// </summary>
+        /// <param name="weaponType">Optional. If not set, the next weapon in the list will be returned (no matter the slot)</param>
+        /// <returns>
+        /// -1: No selection / weapon switch should proceed,
+        /// >0: The index of the weapon in the inventory list that should get selected
+        /// </returns>
```
mzegar, line 57: Nice this is better.

code/player/Player.cs
```diff
@@ -1,16 +1,19 @@
...
 using TTTReborn.UI;
+using TTTReborn.Weapons;
 
 namespace TTTReborn.Player
 {
     public partial class TTTPlayer : Sandbox.Player
     {
         public PlayerCorpse PlayerCorpse { get; set; }
 
+        public static int WeaponDropVelocity { get; set; } = 300;
```
mzegar, line 15: private?

code/player/Player.cs
```diff
@@ -110,6 +117,26 @@ public override void Simulate(Client client)
...
+                if (droppedEntity != null)
+                {
+                    if (droppedEntity.PhysicsGroup != null)
+                    {
+                        droppedEntity.PhysicsGroup.Velocity = Velocity + (EyeRot.Forward + EyeRot.Up) * WeaponDropVelocity;
+                    }
+
+                    timeSinceDropped = 0;
+                }
+            }
```
mzegar, line 134: Do we want to throw this into it's own method? Similar as we've done with

```
TickPlayerUse();
TickAttemptInspectPlayerCorpse();
```

code/ui/WeaponSelection.cs
```diff
@@ -1,37 +1,188 @@
...
+        /// </summary>
+        [Event.BuildInput]
+        public void ProcessClientWeaponSelectionInput(InputBuilder input)
+        {
+            Inventory inventory = Local.Pawn.Inventory as Inventory;
+
+            if (inventory.Count() == 0)
+        	{
+	            return;
+        	}
```
mzegar, line 124: spacing

code/ui/WeaponSelection.cs
```diff
@@ -1,37 +1,188 @@
...
+
+                    input.ActiveChild = inventory.GetSlot(nextSlot);
+
+                    input.MouseWheel = 0;
+                }
+
+                return;
+            }
+
+            // correspondig key pressed
```
mzegar, line 144: spelling

mzegar (commented):
Some more stuff

code/player/Inventory.cs
Alf21: For now, I think private should work as long as we don't need it public

Alf21 (approved):

PR #21, opened 2021-06-13
Hud: Reworked InfoFeed hud element

Alf21:
image

### DIFF
--- a/code/ui/InfoFeed.cs
+++ b/code/ui/InfoFeed.cs
@@ -16,7 +16,7 @@ namespace TTTReborn.UI
             StyleSheet.Load("/ui/InfoFeed.scss");
         }
 
-        public virtual Panel AddEntry(Client leftClient, Client rightClient, string method, string append = null)
+        public virtual Panel AddEntry(Client leftClient, Client rightClient, string method, string postfix = "")
         {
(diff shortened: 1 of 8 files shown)


code/ui/InfoFeed.cs
```diff
@@ -8,20 +13,30 @@ public InfoFeed()
         {
             Current = this;
 
-            StyleSheet.Load( "/ui/InfoFeed.scss" );
+            StyleSheet.Load("/ui/InfoFeed.scss");
         }
 
-        public virtual Panel AddEntry(ulong leftSteamId, string leftName, ulong rightSteamId, string rightName, string method)
+        public virtual Panel AddEntry(Client leftClient, Client rightClient, string method, string append = null)
```
mzegar, line 19: Can we just make this an empty string by default?
`append = ""`

Also, I think we can find a better name for `append`, maybe `postfix`? or `following`

mzegar (approved):
LGTM, one quick comment.

code/ui/InfoFeed.cs
Alf21: If this is better now (ty for the hint, you are right), feel free to resolve and merge :)

PR #24, opened 2021-06-13
Bug: Fixed Prediction with RPCs (Net-Crash) and synced Role manually

Alf21:


code/rounds/InProgressRound.cs
```diff
@@ -128,6 +128,12 @@ private void AssignRoles()
                 {
                     player.Role = new InnocentRole();
                 }
+
+                // send everyone their roles
```
mzegar, line 132: Up to you if you want to leave this, but imo this doesn't provide any more info than the comment already does.

mzegar (approved):
One minor comment, LGTM.

PR #25, opened 2021-06-13
HUD: InspectMenu rework

Alf21:
Fixed continous netstream (replaced by RPC triggers) and improved design.

image

After pressing `E`:
image

### DIFF
--- a/code/player/Player.cs
+++ b/code/player/Player.cs
@@ -17,8 +17,7 @@ namespace TTTReborn.Player
         private static int WeaponDropVelocity { get; set; } = 300;
 
         public BaseRole Role {
-            set
-            {
+            set {
                 role = value;
 
                 if (IsServer)
@@ -40,6 +39,8 @@ namespace TTTReborn.Player
         private float inspectCorpseDistance = 80f;
 
         private TimeSince timeSinceDropped = 0;
+
+        private bool isInspecting = false;
 
         public TTTPlayer()
         {
@@ -142,12 +143,15 @@ namespace TTTReborn.Player
             }
 
             TickPlayerUse();
-            TickAttemptInspectPlayerCorpse();
+
             TickPlayerDropWeapon();
 
             SimulateActiveChild(client, ActiveChild);
 
-            TickAttemptInspectPlayerCorpse();
+            if (IsServer)
+            {
+                TickAttemptInspectPlayerCorpse();
+            }
 
             PawnController controller = GetActiveController();
             controller?.Simulate(client, this, GetActiveAnimator());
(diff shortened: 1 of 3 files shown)


code/player/Player.cs
```diff
@@ -242,14 +255,17 @@ public override void TakeDamage(DamageInfo info)
                 info.Damage *= 2.0f;
             }
 
-            if (info.Attacker is TTTPlayer attacker && attacker != this)
+            using (Prediction.Off())
```
mzegar, line 258: We do not want to turn off prediction here I'm pretty sure. If the FP guys didn't, we shouldn't.

https://wiki.facepunch.com/sbox/Prediction

code/player/Player.cs
```diff
@@ -189,30 +193,39 @@ private void TickPlayerDropWeapon()
...
                     }
 
+                    return;
+                }
+
+                if (inspectingPlayerCorpse != null)
+                {
                     ClientCloseInspectMenu(client);
+
+                    inspectingPlayerCorpse = null;
```
mzegar, line 228: Is there any cases where the inspect menu fails to close?

code/ui/InspectMenu.cs
```diff
@@ -21,33 +22,161 @@ public bool IsShowing
...
-                    + $"{deadPlayer.Role.Name}";
+                confirmationPanel.SetClass("hide", true);
+
+                header.SetPlayer(deadPlayer);
+                header.SetClass("hide", false);
+
+                content.SetPlayer(deadPlayer);
+                content.SetClass("hide", false);
+
+                footer.SetClass("hide", false);
```
mzegar, line 66: Is there a way we can setup this UI to be a single element that gets "hidden"/"unhidden"?

mzegar (commented):
Comments

code/player/Player.cs
Alf21: not in my tests, but feel free trying some edge-cases :D

code/ui/InspectMenu.cs
Alf21: What do you mean exactly? Currently, it's one UI element, that includes more or less 4 panels, 3 of them should just be shown if the corpse was confirmed. The other, if not confirmed. I think that can be handled better as well (just 2 panels), if that's what you meant?

mzegar: If the 3 panels could have one single "hide" call made on them. As in compile the 3 panels into a single "component", then just "hide' the single compontent.

Alf21: Done

mzegar (approved):
LGTM

PR #26, opened 2021-06-13
Shop: Added functional QuickShop

Alf21:
Working version of the shop (without credit usage):
image


Westsaid:
Looks pretty good so far, I'll check this ingame this evening!

Westsaid:
@Alf21 I made some small fixes and enhancements

code/items/weapons/TTTWeapon.cs
```diff
@@ -28,7 +27,7 @@ public WeaponAttribute(string name) : base(name)
     }
 
     [WeaponAttribute("ttt_weapon")]
-    public abstract partial class Weapon : BaseWeapon
+    public abstract partial class TTTWeapon : BaseWeapon, IBuyableItem
     {
         public string Name { get; private set; }
```
mzegar, line 32: Can we get rid of this with `GetName()` now available?

code/player/TTTPlayer.cs
```diff
@@ -74,7 +53,7 @@ public override void Respawn()
             EnableShadowInFirstPerson = true;
 
             Role = new NoneRole();
-            Credits = 0;
+            Credits = 400;
```
mzegar, line 56: just another magic number maybe we want to throw as a server convar.

mzegar (commented):
two quick ones

code/player/TTTPlayer.cs
Alf21: Just for testing, as I don't plan giving every player Credits in the start if we are done.

code/items/weapons/TTTWeapon.cs
Alf21: Good hint, I kept `name` as `private` instead saving performance instead of fetching the name everytime from a class attribute.

mzegar (approved):
Feedback addressed.

PR #27, opened 2021-06-13
Scoreboard: Add groups (Dead)

Alf21:
image


code/player/Player.RPCs.cs
```diff
@@ -34,6 +36,7 @@ public void ClientSetRole(string roleName)
         public void ClientConfirmPlayer(TTTPlayer confirmPlayer, TTTPlayer deadPlayer, string roleName)
         {
             deadPlayer.Role = RoleFunctions.GetRoleByType(RoleFunctions.GetRoleTypeByName(roleName));
+            deadPlayer.IsAlive = false;
```
mzegar, line 39: Instead of `IsAlive` it should be maybe `IsConfirmedDead`, `IsAlive` makes it sounds like it's the state of the player, whereas this seems to be confirming a death.

code/ui/Hud.cs
```diff
@@ -38,9 +45,9 @@ private void OnPlayerDied()
         }
 
         [Event("tttreborn.player.spawned")]
-        private void OnPlayerSpawned()
+        private void OnPlayerSpawned(TTTPlayer player)
         {
-            if (playerInfo != null || weaponSelection != null)
+            if (player != Local.Client.Pawn || playerInfo != null || weaponSelection != null)
```
mzegar, line 50: `player != Local.Client.Pawn`  seems to be an odd check here?

mzegar (approved):
Comments

code/ui/Hud.cs
Alf21: Event `tttreborn.player.spawned` is called for every player. that's why :)

code/player/Player.RPCs.cs
Alf21: That is a good way to additionally keep track of the confirmation state on the server. Feel free modifying the check, currently you can't set IsAlive on server because there is LifeState available (and should be used) :)

Westsaid:
Good step and work! I just added the color of the dead group and changed the text for ScoreboardGroup__title from "Alive (1)" to "Alive - 1". 

I noticed that I  gave the ScoreboardGroup__title a fixed width, We need to improve this earlier or later... Probably need to wrap it in another Panel with flex with another panel as sibling and give them proper flex-grow / flex-shrink attributes...

Shouldn't be too hard, could take over this tomorrow. Do we want to wait with merging or do our further improvements form another branch?

mzegar:
Oops, accidently approved initially. Feel free to ping for any more changes and I'll review regardless.

Westsaid:
Just merged main into this branch with no conflicts. 

image

Adding a wrapper around the scoreboard title fixed the issue.

Alf21:
@mzegar Pls review and merge, if you're okay with it :)

PR #30, opened 2021-06-15
Bug: General fixes

mzegar:
Just some general fixes / changes before we release.

### DIFF
--- a/code/player/camera/SpectateCamera.cs
+++ b/code/player/camera/SpectateCamera.cs
@@ -30,10 +30,9 @@ namespace TTTReborn.Player.Camera
                 return;
             }
 
-            // TODO: Rework spectate camera logic.
             if (TargetPlayer == null || !TargetPlayer.IsValid() || Input.Pressed(InputButton.Attack1))
             {
-                List<TTTPlayer> players = TTTReborn.Gamemode.Game.Instance.Round.Players;
+                List<TTTPlayer> players = GetAlivePlayers();
 
                 if (players.Count > 0)
                 {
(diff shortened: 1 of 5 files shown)


code/player/camera/SpectateCamera.cs
```diff
@@ -71,6 +71,21 @@ private Vector3 GetViewOffset()
...
+            foreach (Client client in Client.All)
+            {
+                if (client.Pawn is TTTPlayer player && player.LifeState == LifeState.Alive)
+                {
+                    players.Add(player);
+                }
+            }
+
+            return players;
+        }
```
Alf21, line 88: Could you put it into Player.cs?
Could you modify <https://github.com/TTTReborn/ttt-reborn/blob/main/code/player/Player.cs#L36-L49> as well to use `Client.All`?
Does it work for client and server? (so Clients know who is alive)?

Alf21 (approved):

code/player/camera/SpectateCamera.cs
Alf21: > Does it work for client and server? (so Clients know who is alive)?

Just tested, indeed, it is... That's bad :D

Alf21: Gonna add this method in the scoreboard PR

mzegar: Are we fine with my changes? I think it'll be better just to start spinning off any util methods we need.

Alf21 (approved):

PR #31, opened 2021-06-17
Scoreboard: Added colored roles and auto confirmation in PostRound

Alf21:
unknown


Westsaid:
One big thing for me: I think we shouldn't color innocents, they are the "normal" players and therefore do not need further attention... this makes the scoreboard too colorful and take away its clean look... If we keep the system like on the web draft we have a self explanatory system: Detes and Traitor have speacial roles, therefore they get a highlight color, every confirmed without a color must be innocent, people will understand this without having the majority of the ScoreboardEntrys green... as Dieter Rams would say "less is more, if less is better". :D

I noticed that when someone else is confirming another dead player its infofeed text color is transparent, this should be white:
image
To test this you can use bots and "bot_mimic 1" and try to identify a dead player with a bot. :D

Also we have to put a special indication on the entry of the own player, I would suggest to give it the yellow color as before when you are innocent, and to give a higher opacity to the bg color if you are traitor/detective.
image


Besides that, everything look good! Hope you agree to my feedback, if not lets discuss it. Keep it up! 🙌

Westsaid (changes requested):
Check comment

Alf21:
Keeping it tracked:

> One big thing for me: I think we shouldn't color innocents, they are the "normal" players and therefore do not need further attention... this makes the scoreboard too colorful and take away its clean look... If we keep the system like on the web draft we have a self explanatory system: Detes and Traitor have speacial roles, therefore they get a highlight color, every confirmed without a color must be innocent, people will understand this without having the majority of the ScoreboardEntrys green... as Dieter Rams would say "less is more, if less is better". :D

This wouldn't support revived players, as they would be placed inside the "alive" group and you maybe wouldn't recognize their role or whether they were innocent or another role (and not confirmed).

> I  noticed that when someone else is confirming another dead player its infofeed text color is transparent, this should be white
Fixed

> Also we have to put a special indication on the entry of the own player, I would suggest to give it the yellow color as before when you are innocent, and to give a higher opacity to the bg color if you are traitor/detective.
This won't support upcoming other roles with the color yellow. Instead, I suggest using a blinking / flashing bar (or with dots instead of colored bars, a colored yellow bar).

code/ui/InfoFeed.cs
```diff
@@ -23,15 +24,19 @@ public virtual Panel AddEntry(Client leftClient, Client rightClient, string meth
             bool isLeftLocal = leftClient == Local.Client;
             bool isRightLocal = rightClient == Local.Client;
 
+            TTTPlayer leftPlayer = leftClient.Pawn as TTTPlayer;
+
             Label leftLabel = e.AddLabel(isLeftLocal ? "You" : leftClient.Name, "left");
-            leftLabel.Style.FontColor = (leftClient.Pawn as TTTPlayer).Role.Color;
             leftLabel.SetClass("me", isLeftLocal);
+            leftLabel.Style.FontColor = leftPlayer.Role is NoneRole ? Color.White : leftPlayer.Role.Color;
```
mzegar, line 31: `NoneRole` has a transparent color in the class, but we use `Color.White` here, any reason why we can't just change it to be `Color.White`? I'm assuming it's probably because it's different colouring we want to use on leaderboard vs. infofeed.

Would it be worth to have to separate colours on each class then?

Just some thoughts.

code/ui/ScoreboardEntry.cs
```diff
@@ -32,7 +41,35 @@ public virtual void UpdateFrom(PlayerScore.Entry entry)
...
+            // TOOD: Make this work on creation of this Entry
+            // TODO add sin-based fading
+            SetClass("me", Local.Client != null && Entry.Get<ulong>("steamid", 0) == Local.Client.SteamId);
+        }
+
+        public override void Tick()
+        {
+            if (client == null)
+            {
+                foreach (Client loopClient in Client.All)
```
mzegar, line 53: eh nitpicky but loopClient just sounds weird... maybe just `c`?

code/ui/ScoreboardEntry.cs
```diff
@@ -32,7 +41,35 @@ public virtual void UpdateFrom(PlayerScore.Entry entry)
...
+            }
+
+            if (client == null || client.Pawn == null || !(client.Pawn is TTTPlayer player) || currentRole == player.Role)
+            {
+                return;
+            }
+
+            currentRole = player.Role;
+
+            RoleColorLabel.Style.BackgroundColor = player.Role is Roles.NoneRole ? player.Role.Color : player.Role.Color.WithAlpha(0.75f);
```
mzegar, line 71: Want to make this alpha colour a static value at the top of the leaderboard?

mzegar (commented):
things to consider

Westsaid (approved):

Westsaid:
Sorry @mzegar @Alf21  didn't see mat commented minutes before I approved and merged this...

PR #32, opened 2021-06-17
Added working Nameplates (one while looking on another player)

Alf21:
image

Just a simple nameplate.

### DIFF
--- a/code/ui/Nameplate.cs
+++ b/code/ui/Nameplate.cs
@@ -6,21 +6,21 @@ using TTTReborn.Player;
 
 namespace TTTReborn.UI
 {
-	public class Nameplate : Panel
-	{
-		public float MaxDrawDistance = 500;
+    public class Nameplate : Panel
+    {
+        public static float MaxDrawDistance = 500;
 
         private Label nameLabel;
 
-		public Nameplate()
-		{
-			StyleSheet.Load("/ui/Nameplate.scss");
+        public Nameplate()
+        {
+            StyleSheet.Load("/ui/Nameplate.scss");
 
             nameLabel = Add.Label("", "name");
-		}
+        }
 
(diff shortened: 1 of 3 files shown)


code/ui/Nameplate.cs
```diff
@@ -0,0 +1,43 @@
...
+using Sandbox.UI;
+using Sandbox.UI.Construct;
+
+using TTTReborn.Player;
+
+namespace TTTReborn.UI
+{
+	public class Nameplate : Panel
+	{
+		public float MaxDrawDistance = 500;
```
mzegar, line 11: this can probably be static private.

code/ui/Nameplate.cs
```diff
@@ -0,0 +1,43 @@
...
+
+        private Label nameLabel;
```
mzegar, line 13: This is indented an extra space in.

code/ui/Nameplate.cs
```diff
@@ -0,0 +1,43 @@
...
+
+		public Nameplate()
+		{
+			StyleSheet.Load("/ui/Nameplate.scss");
+
+            nameLabel = Add.Label("", "name");
```
mzegar, line 19: This is indented an extra space in aswell.

mzegar (approved):
Quick comments, personally, I don't like the black border around the name. It takes up a bit too much of the screen.

We can discuss that at a later point though.

Westsaid:
I agree with mat, I think the border is not necessary in this small panel and can be removed

Westsaid (approved):
Good and simple work!

I got 3 style changes Ill commit afterward, tell me if you are ok with them:
- Remove border 
- Set font-weight : 400
- Half the size of the padding-y

Before: 
image

After:
image

PR #34, opened 2021-06-18
FEA: Post Round Menu

mzegar:
Basic post round menu implementation.

For now just passing around the role name that won, we should later obviously change this to be an object containing all post round data.

image


code/ui/PostRoundMenu.scss
```diff
@@ -0,0 +1,32 @@
...
+    transform: translateX(-50%) translateY(-50%);
+    position: absolute;
+    flex-direction: column;
+
+    color: white;
+    font-family: Poppins;
+    font-weight: 500;
+
+    &.hide {
+        display: none;
```
Westsaid, line 19: Why not opacity:0... this is better in most cases because you can set transitions on opacity but not display:none

code/rounds/InProgressRound.cs
```diff
@@ -141,5 +152,12 @@ private void AssignRoles()
                 }
             }
         }
+
+        private void LoadPostRound(BaseRole winningRole)
+        {
+            // TODO: Consider maybe moving to PostRound
+            TTTPlayer.ClientOpenAndSetPostRoundMenu(winningRole.Name);
```
Westsaid, line 159: Why just the Name? Couldn't we also pass the winningRole.Color to set it as background-color for the Header

code/ui/PostRoundMenu.scss
```diff
@@ -0,0 +1,32 @@
...
+    }
+
+    Header {
+        text-align: center;
+
```
Westsaid, line 24: If we set the BG color, we should also have a `margin:15px 0;`.

Westsaid (changes requested):
Just small design thoughts

Alf21 (approved):

Westsaid:
Good that we can access the Color of the winning team now. I don't really like that the text is colored that way:
image

Its a little hard to read, I would prefer to make it more like a banner: Set the BG color to the winning role. That would look like that:
image
The contrast ratio is more sufficient here.

Westsaid (approved):

PR #35, opened 2021-06-18
Game: Adjusted damage, nameplate tracing and weapon switch shooting delay

Alf21:
This should make the game more "enjoyable" and prevent a scroll-lock shooting bug

### DIFF
--- a/code/weapons/Pistol.cs
+++ b/code/weapons/Pistol.cs
@@ -6,8 +6,8 @@ namespace TTTReborn.Weapons
     partial class Pistol : Weapon
     {
         public override string ViewModelPath => "weapons/rust_pistol/v_rust_pistol.vmdl";
-        public override bool UnlimitedAmmo => false;
-        public override int ClipSize => 10;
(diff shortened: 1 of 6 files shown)


code/weapons/Pistol.cs
```diff
@@ -29,6 +29,11 @@ public override bool CanPrimaryAttack()
 
         public override void AttackPrimary()
         {
+            if (TimeSinceDeployed < 0.4f)
```
mzegar, line 32: Is there any way to use the `deploy` animation instead of just hard coding values?

mzegar (commented):
Comment

code/weapons/Pistol.cs
Alf21: Yea, for sure thought the same before searching these values, but didn't found the right method / var if there was one. Gonna try to add a var to Weapons.cs so we don't need to check it in every weapon manually (maybe found a better way).

mzegar (approved):
LGTM

PR #36, opened 2021-06-19
Teams: Added Traitor syncing

Alf21:
image
image
image


code/player/TTTPlayer.Commands.cs
```diff
@@ -37,6 +41,55 @@ public static void RespawnPlayer(int id)
             }
         }
 
+        [ServerCmd(Name = "requestitem")]
```
mzegar, line 44: `spawnItem` maybe? or did old `GMOD`/`TTT2` use `requestitem`

code/player/TTTPlayer.Confirmation.cs
```diff
@@ -0,0 +1,118 @@
+﻿using Sandbox;
```
mzegar, line 1: File name

`TTTPlayer.PlayerCorpse`? Just nit picking.

code/player/TTTPlayer.Shop.cs
```diff
@@ -0,0 +1,45 @@
...
+using TTTReborn.Items;
+
+namespace TTTReborn.Player
+{
+    public enum BuyError
+    {
+        None,
+        InventoryBlocked,
+        NotEnoughCredits
+    }
```
mzegar, line 12: Neat I like this.

code/rounds/InProgressRound.cs
```diff
@@ -104,20 +106,24 @@ public override void OnPlayerSpawn(TTTPlayer player)
...
+                }
+
+                if (!aliveTeams.Contains(player.Team))
+                {
+                    aliveTeams.Add(player.Team);
+                }
             }
 
-            return new NoneRole();
+            return aliveTeams.Count == 1 ? aliveTeams[0] : null;
```
mzegar, line 126: Are we able to have a `NoneTeam` similar to `NoneRole`? It would just allow us to have a `null`/`empty` state for Team so we can return that instead, just a bit cleaner imo.

code/ui/Nameplate.cs
```diff
@@ -28,16 +29,18 @@ public override void Tick()
                 .Ignore(player)
                 .Run();
 
-            bool targetFound = trace.Hit && trace.Entity is TTTPlayer;
+            bool validHit = false;
```
mzegar, line 32: Could it just be

```csharp
bool validHit = trace.Hit && trace.Entity is TTTPlayer target
if (validHit) {
...
}
SetClass("hide", !validHit);
```

mzegar (commented):
stuff to consider

code/player/TTTPlayer.Commands.cs
Alf21: all commands are lowercase, that's why. But we could talk about `request_item` or `ttt_request_item`

code/player/TTTPlayer.Confirmation.cs
Alf21: Thought the same, but `PlayerCorpse.cs` already exists and could lead to confusion.

code/player/TTTPlayer.Shop.cs
Alf21: For better clientside feedback in the shop later :D

code/rounds/InProgressRound.cs
Alf21: Thought about it, but that could interfer with the win condition. I'd maybe do this after the PR is finished, what do you think?

code/ui/Nameplate.cs
Alf21: Doesn't work sadly, already tried, but if connected `trace.Entity is TTTPlayer` via `&&`, it seems that `target` isn't available anymore.

code/player/TTTPlayer.Confirmation.cs
mzegar: ah right.

code/rounds/InProgressRound.cs
mzegar: Throw in a TODO, we can handle it later.

code/ui/Nameplate.cs
mzegar: huh odd.

code/rounds/InProgressRound.cs
Alf21: Added it into the Trello.

code/ui/Nameplate.cs
Alf21: I threw it into Trello as well so we keep track on it.

mzegar (approved):
Feedback addressed, I'll be stepping out for most of the day, so if this is all you plan to add, feel free to push. Otherwise keep this up.

PR #40, opened 2021-06-20
Meta: Change editorconfig, update codebase

mzegar:
https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/code-style-rule-options

Uses the default options Microsoft provides as a base, with some minor changes.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,8 +1,12 @@
-# EditorConfig is awesome: https://EditorConfig.org
-
-# top-most EditorConfig file
+# Remove the line below if you want to inherit .editorconfig settings from higher directories
 root = true
 
+# C# files
+[*.cs]
+
+#### Core EditorConfig Options ####
+
+# Indentation and spacing
 [*]
 indent_style = space
 indent_size = 4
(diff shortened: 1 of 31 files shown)


code/items/equipments/TTTEquipment.cs
```diff
@@ -16,20 +16,20 @@ public EquipmentAttribute(string name) : base(name)
 
     public abstract class TTTEquipment : Networked, IBuyableItem
     {
-        private string name { get; set; }
+        private string Name { get; set; }
```
mzegar, line 19: All properties are pascal case regardless of access modifier, not sure about the pro vs. cons about handling it this way.

Alf21:
I think that's a PR for @Histalek to review :)

.editorconfig
```diff
@@ -1,8 +1,12 @@
-# EditorConfig is awesome: https://EditorConfig.org
-
-# top-most EditorConfig file
+# Remove the line below if you want to inherit .editorconfig settings from higher directories
 root = true
 
+# C# files
+[*.cs]
```
Histalek, line 5: This needs to be moved below the default `[*]` section, e.g. line 17/18 currently.

Otherwise the csharp and dotnet rules will propably apply to all files instead of only `*.cs` files

.editorconfig
```diff
@@ -11,117 +15,128 @@ charset = utf-8
 trim_trailing_whitespace = true
 insert_final_newline = true
 
-[*.md]
-insert_final_newline = false
-trim_trailing_whitespace = false
```
Histalek, line 16: Please re-add these lines

.editorconfig
```diff
@@ -144,14 +159,56 @@ csharp_space_between_method_declaration_parameter_list_parentheses = false
...
-dotnet_style_parentheses_in_other_operators = never_if_unnecessary
+# Wrapping preferences
+csharp_preserve_single_line_blocks = true
+csharp_preserve_single_line_statements = true
 
-# correct eol for scripts
-[*.sh]
-end_of_line = lf
-[*.{cmd,bat}]
-end_of_line = crlf
```
Histalek, line 157: These lines should probably also be re-added. just to be on the safe side

Histalek (changes requested):
i haven't looked to closely at the style changes, but as long as they lead to a consistent code style i'm fine with it :)

.editorconfig
mzegar: Should already be under [*], 

```
[*]
indent_style = space
indent_size = 4
end_of_line = lf
charset = utf-8
trim_trailing_whitespace = true
insert_final_newline = true
```

Histalek: indeed we want to negate those two settings from the default `[*]` block, because

- two trailing whitespaces equal a linebreak in rendered markdown
- i can't remember right now why we would not want a final newline in markdown files. I hope i added this for a reason :D

Histalek (approved):

PR #41, opened 2021-06-20
Game: Several fixes

Alf21:
Fixes:
- Disabled just player damage in non `InProgressRound`s
- Added player healing on round start in `InProgressRound`
- Changed headshot detection using `HitboxGroup` instead of `HitboxIndex`
- Removed deprecated functions
- Added `player.IsValid()` checks in timers
- Disabled corpse confirming in `PreRound` (`WaitingRound`)
- Added role syncing for new connected players

### DIFF
--- a/code/items/weapons/Knife.cs
+++ b/code/items/weapons/Knife.cs
@@ -61,6 +61,8 @@ namespace TTTReborn.Items
 
         public override void AttackPrimary()
         {
+            (Owner as AnimEntity).SetAnimBool("b_attack", true);
+
             ShootEffects();
             PlaySound("rust_boneknife.attack");
             MeleeStrike(BaseDamage, 1.5f);
@@ -68,6 +70,8 @@ namespace TTTReborn.Items
 
         public override void OnChargeAttackFinish()
         {
+            (Owner as AnimEntity).SetAnimBool("b_attack", true);
+
             ShootEffects();
             PlaySound("rust_boneknife.attack");
             MeleeStrike(BaseDamage * 3f, 1.5f);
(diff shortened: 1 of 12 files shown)


code/items/weapons/Knife.cs
```diff
@@ -61,17 +61,28 @@ public override void AttackSecondary()
 
         public override void AttackPrimary()
         {
-            ShootEffects();
+            (Owner as AnimEntity).SetAnimBool("b_attack", true);
+
+            using (Prediction.Off())
+            {
+                ShootEffects();
```
mzegar, line 68: What's the reason behind turning off prediction here?

My concern would be, since we use prediction for the actual shot on the server, and not on client side, the effects may be out of line.

code/items/weapons/Shotgun.cs
```diff
@@ -55,13 +59,12 @@ protected override void ShootEffects()
             Particles.Create("particles/pistol_ejectbrass.vpcf", EffectEntity, "ejection_point");
 
             ViewModelEntity?.SetAnimBool("fire", true);
+            CrosshairPanel?.CreateEvent("fire");
```
mzegar, line 62: technically isn't needed with the crosshair we currently use.

mzegar (commented):
two quick questions/comments

code/items/weapons/Knife.cs
Alf21: Calling ClientRpc's without turned off predictions, crashes can happen. Until that change, I didn't got any crash anymore, but could be random. `ShootEffects` is a ClientRpc

code/items/weapons/Shotgun.cs
Alf21: That's right, but kept it for consistency and to be able extending it easily later on.

mzegar (approved):
LGTM

PR #42, opened 2021-06-20
UI polish

Westsaid:
Some general changes like really slight adjustments to bg, border & text color.

Polish of the QuickShop styling.
Before:
image

After:
image


code/ui/QuickShop.cs
```diff
@@ -47,20 +47,24 @@ public override void Tick()
...
 
             public Header(Panel parent)
             {
                 Parent = parent;
 
                 TitleLabel = Add.Label("Shop", "title");
-                CreditsLabel = Add.Label("$ 0", "credits");
+                PriceHolder = Add.Panel("priceholder");
+                DollarSign = PriceHolder.Add.Label("$", "dollarsign");
+                CreditsLabel =  PriceHolder.Add.Label("0", "credits");
```
mzegar, line 62: Just a suggestion, up to you, do we want to continue with the `Label` postfix?

`PriceLabel`
`DollarSignLabel`
`CreditsLabel`

feel a bit better to me imo.

code/ui/QuickShop.cs
```diff
@@ -47,20 +47,24 @@ public override void Tick()
 
         private class Header : Panel
         {
+            public Panel PriceHolder;
```
Alf21, line 50:
```suggestion
            public Panel PriceHolder { get; set; }
```

mzegar (approved):
minor comment, up to you if you want to change, approving regardless.

code/ui/QuickShop.cs
```diff
@@ -47,20 +47,24 @@ public override void Tick()
...
             public Label TitleLabel { get; set; }
+            public Label DollarSign { get; set; }
```
Alf21: Do we really need in any situation access to that var? ^^

code/ui/QuickShop.cs
```diff
@@ -155,6 +164,7 @@ public void Update()
                 {
                     IsDisabled = (Local.Pawn as TTTPlayer).CanBuy(buyableItem) != BuyError.None;
 
+
```
Alf21, line 167: ?

code/ui/QuickShop.cs
Westsaid: Agree, Ill update

code/ui/QuickShop.cs
```diff
@@ -204,15 +214,18 @@ public void SetItem(IBuyableItem item)
...
                 public Button BuyButton;
                 public IBuyableItem Item;
 
                 public BuyArea(Panel parent)
                 {
                     Parent = parent;
-
-                    PriceLabel = Add.Label("$ 200", "price");
+                    PriceHolder = Add.Panel("priceholder");
+                    DollarSign = PriceHolder.Add.Label("$","dollarsign");
```
Alf21:
```suggestion
                    DollarSign = PriceHolder.Add.Label("$", "dollarsign");
```

code/ui/QuickShop.scss
```diff
@@ -1,25 +1,24 @@
...
-    background-color: #0009;
-    border: 1px solid #000;
+    background-color: #000c;
+    border: 1px solid #1b1b1b;
     border-radius: 8px;
 
     font-family: Poppins;
-    max-width: 800px;
-    min-width: 600px;
+    width: 600px;
```
Alf21, line 7: I don't like the strict way. With a min-max, you are on the save side regarding rendering, border-sizes etc., as the best fitting size will be taken.

code/ui/QuickShop.scss
```diff
@@ -1,25 +1,24 @@
...
 
-    color: white;
+    color: #fafafa;
     font-family: Poppins;
     font-weight: 500;
     pointer-events: all;
 
     &.hide {
-        display: none;
+        opacity: 0;
```
Alf21, line 21: I think `display` makes more sense because we don't want animations there. If you are a traitor and wanna shop, then the shop shall open immediately.

code/ui/QuickShop.scss
```diff
@@ -34,49 +33,84 @@ QuickShop {
         border-top-right-radius: 7px;
 
         Label {
-            &.title, &.credits {
+            &.title,
+            &.credits {
                 font-size: 20px;
             }
+            &.dollarsign {
```
Alf21, line 40:
```suggestion

            &.dollarsign {
```

code/ui/QuickShop.scss
```diff
@@ -34,49 +33,84 @@ QuickShop {
...
+                margin-top: 4px;
+            }
         }
     }
 
     Content {
+        width: 100%;
```
Alf21, line 47: is this really needed?

code/ui/QuickShop.scss
```diff
@@ -34,49 +33,84 @@ QuickShop {
...
         max-height: 400px;
-        padding: 10px;
+        padding: 16px;
         text-align: center;
-        justify-content: center;
 
         Panel.wrapper {
+            width: 100%;
```
Alf21, line 53: needed?

code/ui/QuickShop.scss
```diff
@@ -34,49 +33,84 @@ QuickShop {
...
+            justify-content: flex-start;
+
             ItemPanel {
                 flex-direction: column;
                 margin: 10px;
+                // margin-right: 50px;
+                padding-right: 13px;
+
+                .priceholder {
+                    display: flex;
```
Alf21, line 64: needed?

code/ui/QuickShop.scss
```diff
@@ -34,49 +33,84 @@ QuickShop {
...
 
         Panel.wrapper {
+            width: 100%;
+            flex-wrap: wrap;
+            justify-content: flex-start;
+
             ItemPanel {
                 flex-direction: column;
                 margin: 10px;
+                // margin-right: 50px;
```
Alf21, line 60: comment?

code/ui/QuickShop.scss
```diff
@@ -34,49 +33,84 @@ QuickShop {
...
+                    Label.price {
+                        font-weight: 400;
+                        font-size: 16px;
+                        margin-right: 1px;
+                    }
 
-                Label.price {
-                    font-weight: 400;
-                    font-size: 16px;
+                    .dollarsign {
```
Alf21, line 74:
```suggestion
                    Label.dollarsign {
```
Or wrap it together with the above one :)

code/ui/QuickShop.scss
```diff
@@ -34,49 +33,84 @@ QuickShop {
...
-                    border: 2px solid transparent;
+                    border: 1px solid #1b1b1b;
                     transition: all 0.05s linear;
 
                     background-size: cover;
                     background-image-tint: white;
                 }
-
-                &:hover, &.active {
+                &:hover {
```
Alf21, line 96:
```suggestion

                &:hover {
```

code/ui/QuickShop.scss
```diff
@@ -34,49 +33,84 @@ QuickShop {
...
                     cursor: pointer;
+                    Panel.icon {
```
Alf21, line 98:
```suggestion

                    Panel.icon {
```

code/ui/QuickShop.scss
```diff
@@ -34,49 +33,84 @@ QuickShop {
...
+                        border: 1px solid #df2935;
+                    }
+                    &.active {
```
Alf21, line 101:
```suggestion

                    &.active {
```

code/ui/QuickShop.scss
```diff
@@ -34,49 +33,84 @@ QuickShop {
...
+                        cursor: default;
 
+                        Panel.icon {
+                            border: 2px solid #df2935;
+                        }
+                    }
+                }
+                &.active {
```
Alf21, line 109:
```suggestion

                &.active {
```

Alf21 (changes requested):

code/ui/QuickShop.cs
mzegar: Yeah a lot of our UI elements just use properties for whatever reason, this PR will fix that, and hopefully when people look back at the files they just convert them to fields.

https://github.com/TTTReborn/ttt-reborn/pull/40

code/ui/QuickShop.scss
Westsaid: We don't necessary need animations here I agree, although I think its better to generally use opacity and have this in any UI element over `display`.

Alf21: Hmm.. What are the reasons for that decision? Doesn't the element still renders with `opacity` while keeping undisplayed with `display` and saving performance? Or does it not matter? Just thought about it at all ^^

code/ui/QuickShop.scss
Westsaid: Ok, I reverted this

code/ui/QuickShop.scss
Westsaid: Idk imo it's more useful to generally use `opacity` over `display` if we want to be able to add short animations later if we decide to so... I also think that a couple of hundred milliseconds wouldn't be a huge blocker a T trying to buy something real quick. I also don't know how big the differences in performance are and if the would make any impact tho... 
But I will revert this back to display...

mzegar:
```
display:none will hide the whole element and remove that from layout space whereas visibility:hidden hides an element but take up the same space as before. Opacity can be used if you want to create transparency or fade effect.
```

Westsaid:
Thanks for your feedback @Alf21 I implemented it. In the process I made a still open bug we still have to fix: The height of the QuickShop content is a big bit bigger than it should be...

Alf21 (approved):
LGTM

PR #45, opened 2021-06-21
Weapon: Added MagnetoStick

Alf21:
unknown
unknown


mzegar:
If you want to checkout https://github.com/Facepunch/sbox-hidden/blob/95b6e6894e8e0994df63bbcfa4a4d56fee2d5cf7/code/player/Player.cs#L217

It does also turn off prediction. (unless you've been already looking at this :D )

code/items/weapons/MagnetoStick.cs
```diff
@@ -0,0 +1,313 @@
...
+using TTTReborn.Player;
+
+// Credits to the PhysGun: https://github.com/Facepunch/sandbox/blob/master/code/tools/PhysGun.cs
+
+namespace TTTReborn.Items
+{
+    [Library("ttt_magnetostick")]
+    partial class MagnetoStick : TTTWeapon // TODO Equipment
+    {
+        public override string ViewModelPath => "weapons/rust_boneknife/v_rust_boneknife.vmdl";
```
mzegar, line 13: Do we want to add a "CanDrop" param? So that users cannot drop this item?

mzegar (commented):
Small comment, nice stuff.

mzegar:
You are also able to move the body completely through the wall, I'm pretty sure the Hidden gamemode handles the ragdoll collision and doesn't let you do this, might be good to check it out.

code/items/weapons/MagnetoStick.cs
Alf21: Nice idea, planned this but forgot somehow :D Gonna do so in this PR

mzegar (approved):
LGTM

Alf21:
> LGTM, what's the update with the bodies moving through the wall, did you manage to fix it?

Yea that's fixed, but sometimes you still get weird physics :D

PR #50, opened 2021-06-23
Fix: (Task) error handling, Hotloading issue with `NetworkedComponent`

Alf21:
Currently, I'm waiting for a decision / feedback regarding https://github.com/Facepunch/sbox-issues/issues/388 as this defines whether we should rework the `Round` system into a manually net-synced system (string based `Round` syncing that survives hotloading) or not.

### DIFF
--- a/code/gamemode/Game.cs
+++ b/code/gamemode/Game.cs
@@ -107,14 +107,12 @@ namespace TTTReborn.Gamemode
                 }
                 catch(Exception e)
                 {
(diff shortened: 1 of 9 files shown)


Alf21:
Okay ready for merge as a hotloading fix (depends on sbox-sided implementation) can be done separately if needed.

mzegar (approved):
As long as it doesn't break things, LGTM

PR #53, opened 2021-06-24
Meta: Setup dotnet linting CI (with GH annotations)

Histalek:


code/ui/WeaponSelection.cs
```diff
@@ -162,7 +162,7 @@ public WeaponSlot(Panel parent, TTTWeapon weapon)
                 Parent = parent;
                 WeaponName = weapon.GetName();
 
-                _slotLabel = Add.Label(((int) weapon.WeaponType).ToString(), "slotlabel");
+                _slotLabel = Add.Label(((int)weapon.WeaponType).ToString(), "slotlabel");
```
Alf21, line 165: imo, there should be a space between type casting (explicit conversions) and the var

Histalek: https://github.com/TTTReborn/ttt-reborn/blob/fea10d705fbdb9b9214753681e98dabd712e922a/.editorconfig#L146

my formatter only did what was specified in the .editorconfig :P
i can change this if the team agrees

saibotk: i would also change it

saibotk (approved):
Awesome. That is exactly how i imagined the check

PR #54, opened 2021-06-25
Game: Add spectator & alive chat/voice

mzegar:
- If we could double check if alive / dead voice chat works, that would be good. There wasn't really a nice way to check it on one machine. But, I'm pretty sure it works.


mzegar:
We also need round checks on both of these... forgot... will fix tomorrow.

code/gamemode/Game.cs
```diff
@@ -85,6 +85,29 @@ public override void ClientDisconnect(Client client, NetworkDisconnectionReason
...
         }
 
+        public override bool CanHearPlayerVoice(Client source, Client dest)
+        {
+            Host.AssertServer();
+
+            var sp = source.Pawn;
+            var dp = dest.Pawn;
+
+            if (sp is not TTTPlayer sourcePlayer || dp is not TTTPlayer destPlayer)
```
Alf21, line 95:
```suggestion
            if (source.Pawn is not TTTPlayer sourcePlayer || dest.Pawn is not TTTPlayer destPlayer)
```

code/ui/ChatBox.cs
```diff
@@ -0,0 +1,137 @@
...
+    public static partial class Chat
+    {
+        public static event Action OnOpenChat;
+
+        [ClientCmd("openchat")]
+        internal static void MessageMode()
+        {
+            OnOpenChat?.Invoke();
+        }
+
```
Alf21, line 19:
```suggestion
```

code/ui/ChatBox.cs
```diff
@@ -0,0 +1,137 @@
...
+            {
+                return;
+            }
+
+            var msg = _input.Text.Trim();
+            _input.Text = "";
+
+            if (string.IsNullOrWhiteSpace(msg))
+                return;
+
```
Alf21, line 75:
```suggestion
            if (string.IsNullOrWhiteSpace(msg))
            {
                return;
            }
```

code/ui/ChatBox.cs
```diff
@@ -0,0 +1,137 @@
...
+
+            Say(msg, player.LifeState);
+        }
+
+        public void AddEntry(string name, string message, string avatar, LifeState lifeState)
```
Alf21, line 80: Maybe we should pass the class name directly as it would be nice if traitors would see other traitors as well. Or an option adding additional classes.

code/ui/ChatBox.cs
```diff
@@ -0,0 +1,137 @@
...
+
+        [ServerCmd("say")]
+        public static void Say(string message, LifeState lifeState)
+        {
+            Assert.NotNull(ConsoleSystem.Caller);
+
+            // TODO: Reject more messed up user inputs
+            if (message.Contains('\n') || message.Contains('\r'))
+            {
+
```
Alf21, line 120: ?

code/ui/ChatBox.cs
```diff
@@ -0,0 +1,137 @@
...
+            }
+
+            Log.Info($"{ConsoleSystem.Caller}: {message}");
+
+            if (Gamemode.Game.Instance?.Round is Rounds.InProgressRound && lifeState == LifeState.Dead)
+            {
+                var deadClients = Gamemode.Game.GetDeadClients();
+                AddChatEntry(To.Multiple(deadClients), ConsoleSystem.Caller.Name, message, $"avatar:{ConsoleSystem.Caller.SteamId}", lifeState);
```
Alf21, line 128:
```suggestion
                AddChatEntry(To.Multiple(Gamemode.Game.GetDeadClients()), ConsoleSystem.Caller.Name, message, $"avatar:{ConsoleSystem.Caller.SteamId}", lifeState);
```

code/ui/ChatEntry.cs
```diff
@@ -0,0 +1,33 @@
+using Sandbox;
+using Sandbox.UI;
+using Sandbox.UI.Construct;
+
```
Alf21, line 4:
```suggestion
```

Alf21 (changes requested):

Alf21 (approved):

PR #57, opened 2021-06-25
Meta: Add contributing guidelines and issue template

Histalek:


.github/ISSUE_TEMPLATE/config.yml
```diff
@@ -0,0 +1,5 @@
+blank_issues_enabled: false
```
Histalek, line 1: We need to check how this works out with us using GitHub-Projects.
On the contrary defining a project-task template might not be the worst of ideas.

CONTRIBUTING.md
```diff
@@ -0,0 +1,74 @@
+# ![TTTReborn logo](https://github.com/TTTReborn/graphics/raw/main/github_social/social.png) Contributing to TTTReborn
```
Histalek, line 1: looking at the rich diff, the banner might be a bit too much. Thoughts?

saibotk: You can use html markup to reduce the size and even center it etc.

Also you might want to change the image to another one that is only the logo

saibotk (approved):

PR #71, opened 2021-06-25
Scoreboard: Added  `MIA` group

Alf21:
image

The `MIA` is just shown for `Traitors` team currently.

### DIFF
--- a/code/player/TTTPlayer.Confirmation.cs
+++ b/code/player/TTTPlayer.Confirmation.cs
@@ -11,6 +11,8 @@ namespace TTTReborn.Player
 
         public bool IsConfirmed = false;
 
+        public bool IsMissingInAction = false;
+
         public TTTPlayer CorpseConfirmer = null;
 
         private const float INSPECT_CORPSE_DISTANCE = 80f;
--- a/code/ui/Scoreboard.cs
+++ b/code/ui/Scoreboard.cs
@@ -183,7 +183,7 @@ namespace TTTReborn.UI
                     {
                         if (client.Pawn.IsValid() && client.SteamId == steamId)
                         {
-                            if((client.Pawn as TTTPlayer).IsMissingInAction)
+                            if ((client.Pawn as TTTPlayer).IsMissingInAction)
                             {
                                 group = "MIA";
                             }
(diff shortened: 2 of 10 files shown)


code/player/TTTPlayer.Confirmation.cs
```diff
@@ -11,6 +11,8 @@ partial class TTTPlayer
 
         public bool IsConfirmed = false;
 
+        public bool IsMissingInAction = false;
```
mzegar, line 14: Should we just have an enum for the state of the player? instead of having separate bools for each?

i.e
```csharp
enum PlayerStatus {
	Confirmed
	MIA
	Alive
	Dead
}
```

code/ui/Scoreboard.cs
```diff
@@ -175,6 +173,26 @@ private ScoreboardGroup GetScoreboardGroup(PlayerScore.Entry entry)
...
+
+                if (steamId != 0)
+                {
+                    foreach (Client client in Client.All)
+                    {
+                        if (client.Pawn.IsValid() && client.SteamId == steamId)
+                        {
+                            if ((client.Pawn as TTTPlayer).IsMissingInAction)
+                            {
+                                group = "MIA";
```
mzegar, line 188: We might want to make these static strings at the top of the file, or, if we decide with the enum change I proposed, you can just toString() the enum value.

mzegar (commented):
Quick comments, would like to hear your thoughts.

code/player/TTTPlayer.Confirmation.cs
Alf21: Sadly, that doesn't work as a dead player can be confirmed. And an alive player can be alive and MIA.

mzegar: How many possible combinations are there? There is the possibility of either...

1. Just having all the possible combinations in the enums
2. Use this https://stackoverflow.com/questions/19582477/combining-enum-values-with-bit-flags

Alf21: Hmmm..  maybe for another PR. It's not as simple keeping track of them properly as Alive and Dead arise from the LifeState which is handled automatically and is synced to all clients as well.

mzegar: kk throw up a TODO / task and I'll check it out.

mzegar (approved):
Just add an issue onto github and I'll check it out, LGTM.

PR #79, opened 2021-06-26
Rounds: Fixed that `WaitingRound` never triggered after a round started

Alf21:
Before this change, there never have been a jump back to the `WaitingRound` if too many player disconnected and the minimum amount of players wasn't reached.

I changed it into the following behavior (if there are too few players):
- `PreRound`: Run the countdown until it reaches the end, then do a check for enough players (So if a player connects and disconnects, the whole preparing isn't started from the beginning every time).
- `InProgressRound`: Check every second for enough players. If there aren't enough (and still no winning team), switch to `WaitingRound`
- `PostRound`: Same as `PreRound`

### DIFF
--- a/code/rounds/InProgressRound.cs
+++ b/code/rounds/InProgressRound.cs
@@ -173,5 +173,25 @@ namespace TTTReborn.Rounds
                 winningTeam.Color
             );
         }
+
+        private bool CheckMinimumPlayers()
+        {
+            return Client.All.Count >= TTTReborn.Gamemode.Game.TTTMinPlayers;
+        }
+
+        public override void OnSecond()
+        {
+            if (Host.IsServer)
+            {
+                if (CheckMinimumPlayers())
+                {
+                    base.OnSecond();
+                }
+                else if (IsRoundOver() == null)
+                {
+                    Gamemode.Game.Instance.ChangeRound(new WaitingRound());
+                }
+            }
+        }
     }
 }
--- a/code/rounds/InProgressRound.cs
+++ b/code/rounds/InProgressRound.cs
@@ -11,6 +11,8 @@ using TTTReborn.Teams;
 
 namespace TTTReborn.Rounds
 {
+    using Gamemode;
+
     public class InProgressRound : BaseRound
     {
         public override string RoundName => "In Progress";
@@ -167,7 +169,7 @@ namespace TTTReborn.Rounds
 
         private static void LoadPostRound(TTTTeam winningTeam)
         {
-            Gamemode.Game.Instance.ChangeRound(new PostRound());
+            Gamemode.Game.Instance.ForceRoundChange(new PostRound());
             TTTPlayer.ClientOpenAndSetPostRoundMenu(
                 winningTeam.Name,
                 winningTeam.Color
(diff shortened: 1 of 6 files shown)


code/rounds/InProgressRound.cs
```diff
@@ -173,5 +173,25 @@ private static void LoadPostRound(TTTTeam winningTeam)
                 winningTeam.Color
             );
         }
+
+        private bool CheckMinimumPlayers()
```
mzegar, line 177: Can we put this in the base class? or does it work as a `static` method somewhere?

code/rounds/PostRound.cs
```diff
@@ -60,5 +64,17 @@ protected override void OnStart()
...
+        {
+            if (Client.All.Count < TTTReborn.Gamemode.Game.TTTMinPlayers)
+            {
+                TTTReborn.Gamemode.Game.Instance.ChangeRound(new WaitingRound());
+
+                return false;
+            }
+
+            return true;
+        }
```
mzegar, line 78: This method should be named appropriate for what it does

If I were to read it, I would assume that it would just check for minimum players, but it also ends up changing the round.

We should split this functionality up into two different methods imo.

code/rounds/PreRound.cs
```diff
@@ -69,5 +72,17 @@ public override void OnPlayerSpawn(TTTPlayer player)
...
+        {
+            if (Client.All.Count < TTTReborn.Gamemode.Game.TTTMinPlayers)
+            {
+                TTTReborn.Gamemode.Game.Instance.ChangeRound(new WaitingRound());
+
+                return false;
+            }
+
+            return true;
+        }
```
mzegar, line 86: We also have just duplicate methods redefined, is there a place where we could put these functions? maybe in `BaseRound`? Maybe as a static method?

mzegar (changes requested):
Good catch on these errors, I personally think a function name should reflect what it actually ends up doing.

code/rounds/InProgressRound.cs
Alf21: If you look into every round, the functions are different from another, afaik just reused one time. But yea, we could put it somewhere, just thought it should be handled individually.

code/rounds/PostRound.cs
Alf21: Yea good idea, any suggestions? :D

mzegar: Let me think on that

code/rounds/InProgressRound.cs
AMastryukov: This could be a static property I think

code/rounds/PostRound.cs
AMastryukov: We could use a base class property "HasEnoughPlayers" and a "CheckHasEnoughPlayers" method that does the round switching logic based on the property.

code/rounds/PreRound.cs
AMastryukov: BaseRound would be my preferred way of checking.

mzegar (approved):
Just merge if you are fine with this since you originally created the PR @Alf21

PR #83, opened 2021-06-27
Hotloading: Added hotloading support to the HUD

Alf21:
Added hotloading support for HUD panels (they are getting rebuild on hotload)


mzegar:
@Alf21 fine to close this?

PR #84, opened 2021-06-27
Feature: Flashlight added

Alf21:
Currently, the flashlight is still laggy.

Already tried adding a KeyframedEntity, but that seems a bit laggy. Tried setting the position / rotation every Tick (Simulate), but that was as well laggy.
Only in-sync entity is the ViewModel of a weapon currently, but that isn't available all the time (e.g. on weapon switch or if the player does not have any weapons equipped). And yea, I tried terry's eye_L bone as well (btw. eye_R acts weird ^^), but this also includes blink and eye idle animations, so not the best solution (but closest solution maybe).

Best way would be to have an entity or bone that is synced with the player's view (EyeRot) all the time (predicted) and parenting it with the flashlight (`SetParent()`).

Current solution: Created a KeyframedEntity (parented with the client's flashlight) which updates the position via `TryKeyframeTo`.


code/player/TTTPlayer.Flashlight.cs
```diff
@@ -0,0 +1,172 @@
...
+            if (shouldShow)
+            {
+                if (!HasFlashlightEntity)
+                {
+                    if (IsServer)
+                    {
+                        _worldFlashlight = new Flashlight();
+                        _worldFlashlight.EnableHideInFirstPerson = true;
+                        _worldFlashlight.Rotation = EyeRot;
+                        _worldFlashlight.Position = EyePos + EyeRot.Forward * 40f;
```
mzegar, line 77: Can we explain these magic numbers? i.e 40f? `const` them and name them based on what they are doing.

code/player/TTTPlayer.Flashlight.cs
```diff
@@ -0,0 +1,172 @@
...
+                        _worldFlashlight.SetParent(this);
+                    }
+                    else
+                    {
+                        _flashlightHolder = new KeyframeEntity();
+                        _flashlightHolder.Position = EyePos + EyeRot.Forward * 40f;
+                        _flashlightHolder.Rotation = EyeRot;
+                        //_flashlightHolder.SetParent(this);
```
mzegar, line 85: random comment here.

code/player/TTTPlayer.Flashlight.cs
```diff
@@ -0,0 +1,172 @@
...
+                        _viewFlashlight.Position = EyePos + EyeRot.Forward * 40f;
+                        _viewFlashlight.Rotation = EyeRot;
+                        _viewFlashlight.SetParent(_flashlightHolder);
+                    }
+                }
+                else
+                {
+                    if (IsServer)
+                    {
+                        // TODO: This is a weird hack to make sure the rotation is right.
```
mzegar, line 98: This doesn't seem like much of a TODO, what's the goal in the future here?

mzegar (changes requested):
If you kill a player that has the flashlight on it causes an exception, might be good to check some other cases as well.

code/player/TTTPlayer.Flashlight.cs
Alf21: To remove this if it was fixed ^^

code/player/TTTPlayer.Flashlight.cs
Alf21: Not exactly. This would be the best way, but it doesn't support EyeRot synced positioning. That's why I kept it ^^

mzegar: I think we should add why it's there then : D

mzegar (approved):

PR #85, opened 2021-06-27
Nameplate health indication & role display

Westsaid:
image
image

To discuss: How do we want to group the damage indicator?

health>70 : "Healthy"
health>20 : "Injured"
else : "Near death"
Whats your opinion?

TODO: 
- _Color the damageIndicator label._ **DONE**
- _BUG: The color of the damageIndicator label only changes when you let the nameplate close & open again._ **DONE**
- _Remove the role coloring of the Nameplate bg to ensure contrast ratio to the color of the damageIndicator label.
  I thought about adding a colored dot before the playername like on the scoreboard._ **DONE**


Alf21:
I like what you did, but idk about the change of the role displaying <https://user-images.githubusercontent.com/58175915/123549032-90732280-d767-11eb-928e-19bb226d9d23.png>, maybe we need something like a team indicator so you can easily and fast detect a teammember :)

Alf21:
And we need a damage overlay if you getting hit (what I initially meant with "DamageIndicator" :D) like seen here <https://youtu.be/sLgqtRUuYB4?t=30>

Westsaid:
@Alf21 I agree it might be a bit hard to quickly identify a T-mate... The best way to achieve this is imo to implement a 3D UI element which is floating over him, in that case you would also be able to see him far across the map. 

Meanwhile speaking of the nameplate here are all the options I currently have in my mind:
- Bg in role colors just like before: Hardly compatible with the colored health text...
- Dot like currently
- Colored 2px borders: 
image
-  Colored name underline
image

Alf21:
Hmmm... Idk tbh, but if I had to choose, I'd take the border solution

mzegar:
Not sure if I'm in the minority, but I'd prefer removing the black transparent square and make it similar to TTT. I like the dot, and the coloured damage indicator label. Thoughts?

Westsaid:
@mzegar We could try and see... the bg is to ensure contrast and the ability to easily read the text...

TimGoll:
In TTT2 I added a colored role icon next to the player name

Westsaid:
I've implemented the bordered version and did some adjustments, this is ready from my pov.
image

code/player/TTTPlayer.Lookup.cs
```diff
@@ -0,0 +1,50 @@
...
+                    // Send the request to the player looking at the player corpse.
+                    // https://wiki.facepunch.com/sbox/RPCs#targetingplayers
+
+                    if (!player.IsValid())
+                    {
+                        return;
+                    }
+
+                    SetNameplateHealth(client, player.Health);
+                    return;
```
Alf21, line 29:
```suggestion
```

code/player/TTTPlayer.Lookup.cs
```diff
@@ -0,0 +1,50 @@
...
+            {
+                To client = To.Single(this);
+                TTTPlayer player = IsLookingAtPlayer();
+
+                if (player != null)
+                {
+                    // Send the request to the player looking at the player corpse.
+                    // https://wiki.facepunch.com/sbox/RPCs#targetingplayers
+
+                    if (!player.IsValid())
```
Alf21, line 23: This can be merged with just `if player.IsValid()` afaik

code/player/TTTPlayer.Lookup.cs
```diff
@@ -0,0 +1,50 @@
...
+    {
+        public TTTPlayer Player { get; set; }
+
+        private const float MAX_DRAW_DISTANCE = 500;
+
+        private void TickAttemptLookup()
+        {
+            using (Prediction.Off())
+            {
+                To client = To.Single(this);
```
Alf21, line 15: Declaration isn't needed and can be used inline.

code/player/TTTPlayer.RPCs.cs
```diff
@@ -114,6 +114,12 @@ public static void ClientCloseInspectMenu()
             }
         }
 
+        [ClientRpc]
+        public static void SetNameplateHealth(float health)
```
Alf21, line 118:
```suggestion
        public static void ClientSetNameplateHealth(float health)
```

code/ui/Nameplate.cs
```diff
@@ -9,15 +9,68 @@ namespace TTTReborn.UI
...
+            return health > 70 ? "Healthy"
+                : health > 20 ? "Injured"
+                : "Near death";
+        }
+
+        private static Color GetHealthColor(float health)
+        {
+            Color healthy = Color.FromBytes(44, 233, 44);
+            Color injured = Color.FromBytes(233, 135, 44);
+            Color near_death = Color.FromBytes(252, 42, 42);
```
Alf21, line 64: Maybe we should put this into a struct or smth like this

code/player/TTTPlayer.Lookup.cs
```diff
@@ -0,0 +1,50 @@
...
+                {
+                    // Send the request to the player looking at the player corpse.
+                    // https://wiki.facepunch.com/sbox/RPCs#targetingplayers
+
+                    if (!player.IsValid())
+                    {
+                        return;
+                    }
+
+                    SetNameplateHealth(client, player.Health);
```
Alf21, line 28: If we would use a struct like suggested, we don't need to send the full exact health but could instead just send a string as a health indicator (used to fetch the right state). But as AC doesn't care as much, we don't need this.

Alf21 (changes requested):

code/ui/Nameplate.cs
```diff
@@ -9,15 +9,68 @@ namespace TTTReborn.UI
...
+
+            return health > 70 ? healthy
+                : health > 20 ? injured
+                : near_death;
```
mzegar, line 68: Should const these values up above for easy access `70`, `20`.

code/ui/Nameplate.cs
```diff
@@ -9,15 +9,68 @@ namespace TTTReborn.UI
...
+
+            _damageIndicatorLabel = _labelHolder.Add.Label("", "damageIndicator");
+
+        }
+
+        private string GetHealthGroup(float health)
+        {
+            return health > 70 ? "Healthy"
+                : health > 20 ? "Injured"
+                : "Near death";
```
mzegar, line 57: Same as other comment, we should const these magic numbers. And even potentially the strings themself.

mzegar (changes requested):
I've already expressed my opinion on how I feel about the large black background, so if everyone else is likes it, go with it 👍 .

Personally, I hate having a large UI pop up near the crosshair.

Alf21 (approved):

mzegar (approved):

PR #86, opened 2021-06-27
Hud: Rework hud, add hot reload to hud

mzegar:


code/ui/Hud.cs
```diff
@@ -23,54 +19,100 @@ public Hud()
...
-            RootPanel.AddChild<QuickShop>();
+            GeneralHudPanel = new GeneralHud(RootPanel);
+            AliveHudPanel = new AliveHud(RootPanel);
+        }
 
-            Scoreboard = RootPanel.AddChild<Scoreboard>();
+        [Event.Hotload]
+        public static void OnHotReloaded()
+        {
+            if (Host.IsClient)
```
Alf21, line 31: is `Host.` really needed?

code/ui/Hud.cs
```diff
@@ -23,54 +19,100 @@ public Hud()
...
+            {
+                Local.Hud?.Delete();
```
Alf21, line 33: What is this? :o I'd just use `Current.ClearChildren(true)`

Alf21: Then you don't have to destroy the whole hud and reinitialize a new object

code/ui/Hud.cs
```diff
@@ -23,54 +19,100 @@ public Hud()
...
+        }
 
-            _inspectMenu ??= RootPanel.AddChild<InspectMenu>();
+        public class AliveHud : Panel
+        {
+            private PlayerInfo _playerInfo;
+            private WeaponSelection _weaponSelection;
+            private InspectMenu _inspectMenu;
+            private Nameplate _nameplate;
+            private QuickShop _quickShop;
```
Alf21, line 87: if we don't need to access any of them, we could put them in a list to avoid redundant code.

Alf21 (changes requested):

code/ui/Hud.cs
mzegar: Yes, since we are in a static method.

code/ui/Hud.cs
mzegar: If we are going to do `new Hud();` it probably makes sense to just delete the entire object. Since on the next line we are constructing a brand new hud.

Alf21 (approved):

PR #88, opened 2021-06-27
Bug: inspecting menu showing preround

mzegar:
I personally feel like we shouldn't be doing these types of checks, I'm pretty sure in regular TTT inspecting works always, the other issue is that we potentially run into bugs like these.

Bodies also delete pre-round, so inspecting these pre-round shouldn't be an issue.


Alf21:
Hmm idk tbh, because

> Bodies also delete pre-round

just after ~1s, so you could trigger a confirmation.

Alf21 (approved):

PR #90, opened 2021-07-03
Inventory: Added `Ammo` and `Perks`

Alf21:


code/player/Inventory.Perks.cs
```diff
@@ -0,0 +1,83 @@
...
+            }
+
+            return true;
+        }
+
+        public bool Take(TTTPerk perk)
+        {
+            if (!Has(perk))
+            {
+                return false;
```
mzegar, line 44: Just a minor thing to consider, I see "issues" making some of these methods return a `bool` since we don't exactly know what the return type means. 

For example, I think most would assume that `Take` would return `True` if the player no longer has the perk on them. However, in this case, we return `False` if the perk removal fails.

Something to consider.

mzegar (approved):
I would personally reconsider some of these methods and the meaning behind the returning of a `bool`. Nothing to hold this PR back for though, LGTM.

PR #93, opened 2021-07-04
Hud: Added perks to the UI

Alf21:
image


code/items/perks/Radar1.cs
```diff
@@ -0,0 +1,15 @@
+using Sandbox;
+
+namespace TTTReborn.Items
+{
+    [Library("ttt_radar1")]
+    public class Radar1 : TTTPerk, IBuyableItem
```
mzegar, line 6: What's the difference between `Radar1` vs. `Radar2`?

code/ui/Hud.cs
```diff
@@ -34,10 +33,14 @@ public static void OnHotReloaded()
...
             {
                 Local.Hud?.Delete();
+
                 Hud hud = new Hud();
+
                 if (Local.Client.Pawn is TTTPlayer player && player.LifeState == LifeState.Alive)
                 {
                     hud.AliveHudPanel.CreateHud();
+
+                    hud.AliveHudPanel.Effects.OnHotReloaded();
```
mzegar, line 43: Can't we create the elements inside of `CreateHud()`?

code/ui/Hud.cs
```diff
@@ -95,13 +100,16 @@ public void CreateHud()
             {
                 if (_panels.Count == 0)
                 {
+                    Effects = Parent.AddChild<Effects>();
```
mzegar, line 103: If we remove the earlier `Effects.OnHotReloaded` then we don't have to uniquely handle `Effects` here right?

mzegar (changes requested):
If you need to work off a branch, you can just make a branch of a branch. This would be helpful since I had to skim through a lot of this stuff in a different PR. 

https://stackoverflow.com/questions/4470523/create-a-branch-in-git-from-another-branch

42 files is a lot : D

code/items/perks/Radar1.cs
Alf21: Just testing the UI stuff

code/ui/Hud.cs
Alf21: We don't know whether it's hotloaded then :o

code/ui/Hud.cs
Alf21: We need this because it's e.g. used in `TTTPerks.OnEquip`

mzegar (approved):
Made some changes, check em over.

code/ui/Effects.cs
```diff
@@ -8,45 +8,40 @@
...
+        public void AddEffect(TTTPerk perk)
         {
-            PerksInventory perks = ((Local.Pawn as TTTPlayer).Inventory as Inventory).Perks;
+            _effectList.Add(new Effect(this) {item = perk});
+        }
 
-            for (int i = 0; i < perks.Count(); i++)
+        public void RemoveEffect(TTTPerk perk)
+        {
+            foreach (Effect effect in _effectList.Where(effect => effect.item.Name == perk.Name))
```
Alf21: Just one item of the same name can exist in a list, so we don't need a foreach loop theoretically. Maybe merge everything using Linq? :)

PR #96, opened 2021-07-04
Fix: Fixed weapon spawning when adding to the inventory failed

Alf21:
This does not fix the weapon spawning issue when the game hotloaded. It's caused by the shop item's creation and just can be fixed if we are able to use custom LibraryAttributes on weapons. Current workaround is to still spawn the weapons and just disable drawing (invisible weapons).

Thought about copying the data into a struct and then remove the created item again, but that would still spawn a weapon for a short time, so I just wanna wait for the fix sbox-sided.

### DIFF
--- a/code/items/IBuyableItem.cs
+++ b/code/items/IBuyableItem.cs
@@ -10,9 +10,10 @@ namespace TTTReborn.Items
 
         void OnPurchase(TTTPlayer player)
         {
-            (player.Inventory as Inventory).Add(this);
-
-            Equip(player);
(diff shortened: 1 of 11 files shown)


mzegar:
Someone else can voice their opinion, but I feel as if we should just wait instead of putting in a half-fix for this issue.

Regardless, thanks for checking this out. Does an issue have to be created on sbox-issues?

Alf21:
Already created on 2 weeks ago :D ref. <https://github.com/Facepunch/sbox-issues/issues/376>

mzegar (approved):
since nobody else approved, :D

PR #100, opened 2021-07-08
Fea: Improve spectating, add free camera

mzegar:
Players start in a ragdoll spectating state, if they hit space they'll be taken to a free camera, then to a third person camera, then back to free camera.


mzegar:
I wanted to put this logic inside of `TTTPlayer` since we will at some point need to handle processing props, which will be easier inside of here.

mzegar:
Marking as draft, have another minor improvement I want to make tomorrow.

mzegar:
I'm closing this for now, I want to think over this and wait for the spectator changes to occur.

PR #104, opened 2021-07-10
MagnetoStick: Fix rope behavior (not visual)

AMastryukov:
- Add some prevention against prop killing
- Fix grabbing distance naming
- Fix player corpse rope spring naming

### DIFF
--- a/code/items/equipments/MagnetoStick.cs
+++ b/code/items/equipments/MagnetoStick.cs
@@ -12,7 +12,8 @@ namespace TTTReborn.Items
     {
         public override string ViewModelPath => "";
 
-        private static int _grappingDistance => 80;
+        private static int _grabbingDistance => 80;
+        private static float _maxPropSpeed => 10f;
 
         private PhysicsBody _holdBody;
         private WeldJoint _holdJoint;
@@ -22,6 +23,8 @@ namespace TTTReborn.Items
 
         private float _holdDistance = 0f;
         private bool _isAttaching = false;
+
+        private Vector3 _previousPosition;
 
         public Entity GrabbedEntity { get; set; }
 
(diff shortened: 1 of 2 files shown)


mzegar:
Some notes

1. Bodies fall out of your magneto stick if you change your direction really quickly. Not sure if we want to keep this / fix this. (This happens even if you are holding left click)
2. Ropes behave a lot better, which is awesome.

I think in the OG TTT you cannot give any velocity to bodies, only props. Do we want to follow this as well?

AMastryukov:
> Some questions/comments
> 
> 1. Bodies fall out of your magneto stick if you change your direction really quickly. Not sure if we want to keep this / fix this. (This happens even if you are holding left click)
> 2. Ropes behave a lot better, which is awesome.
> 
> I think in the OG TTT you cannot give any velocity to bodies, only props. Do we want to follow this as well?

1. This is something we might want to figure something out about this. The problem I tried to resolve was prop killing, which is actually far easier to do here than in GMod TTT, especially because you can hold a prop and just move it and it will kill people with the force. We could put an if statement that ignores ragdolls perhaps?
2. I have tried fixing the rope particle but it seems to be totally broken.

code/items/equipments/MagnetoStick.cs
```diff
@@ -165,11 +169,19 @@ private void GrabMove(Vector3 startPos, Vector3 dir, Rotation rot)
 
             _holdBody.Position = startPos + dir * _holdDistance;
             _holdBody.Rotation = rot * _heldRot;
+
+            if (Vector3.DistanceBetween(_previousPosition, _heldBody.Position) > _maxPropSpeed)
+            {
+                GrabEnd();
+                return;
```
Alf21, line 176:
```suggestion

                return;
```

mzegar (approved):
LGTM, get @Alf21 to approve as well since he wrote it, he knows this stuff better.

Alf21 (commented):
The problem I've with physicjoint springs is that they act weird often, as you can see here:
image

As a result (the distance difference), you can't pickup the body anymore.

PR #108, opened 2021-07-11
Shop: Fixed weapon spawning

Alf21:
Shop does not require created objects anymore (just for initialization)


mzegar (approved):
LGTM, don't have much to say on my end.

PR #109, opened 2021-07-11
Fix: inventory selecting UI getting cleared

mzegar:
Base.DeleteContents already clears the list.

### DIFF
--- a/code/player/Inventory.cs
+++ b/code/player/Inventory.cs
@@ -21,20 +21,13 @@ namespace TTTReborn.Player
 
         public override void DeleteContents()
         {
-            foreach (Entity entity in List)
-            {
-                if (entity is IItem item)
-                {
-                    item.Remove();
-                }
-            }
-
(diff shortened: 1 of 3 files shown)


code/player/Inventory.cs
```diff
@@ -21,20 +21,13 @@ public Inventory(TTTPlayer player) : base(player)
 
         public override void DeleteContents()
         {
-            foreach (Entity entity in List)
-            {
-                if (entity is IItem item)
-                {
-                    item.Remove();
-                }
-            }
```
Alf21, line 30: This is needed to call the `OnRemove` method for weapons as well

mzegar: ah

Alf21: Haha yea, such confusing, but to keep it in line with the other Items.

Alf21 (approved):

PR #111, opened 2021-07-11
Damage: Added drowning damage, improved falldamage (velocity based), added DamageIndicator, improved RPCs

Alf21:
image

image


code/player/controller/DefaultWalkController.cs
```diff
@@ -12,6 +12,15 @@ public partial class DefaultWalkController : WalkController
         public float MaxSprintSpeed = 300f;
         public float StaminaLossPerSecond = 30f;
         public float StaminaGainPerSecond = 25f;
+        public float FallDamageVelocity = 550f;
+        public float FallDamageScale = 0.25f;
+        public bool IsDiving = false;
+        public float DrownDamageTime = 15f;
+        public float DrownDamagePerSecond = 10f;
+        public TimeSince TimeSinceDivingStarted = 0f;
```
mzegar, line 20: Just suggesting maybe some potential renames

```csharp
IsDiving => IsUnderwater // To avoid confusion with the action of diving itself.
DrownDamageTime => DrowningDuration? // This one is a bit more difficult.
TimeSinceDivingStarted => TimeSpentUnderwater/TimeUnderwater/DurationUnderwater
```

code/player/controller/DefaultWalkController.cs
```diff
@@ -43,9 +52,75 @@ public override void Simulate()
                 SprintSpeed = (MaxSprintSpeed - DefaultSpeed) / player.MaxStamina * player.Stamina + DefaultSpeed;
             }
 
+            IsDiving = Pawn.WaterLevel.Fraction == 1f;
+
+            if (!IsDiving)
+            {
+                TimeSinceDivingStarted = 0f;
+            }
```
mzegar, line 60: Do we want to do these checks on server? It might be better to see if the player is underwater on server as opposed to client.

code/player/controller/DefaultWalkController.cs
```diff
@@ -43,9 +52,75 @@ public override void Simulate()
...
+                    {
+                        Attacker = Pawn,
+                        Flags = DamageFlags.Drown,
+                        HitboxIndex = (int) HitboxIndex.Head,
+                        Position = Position,
+                        Damage = MathF.Max(DrownDamagePerSecond * Time.Delta, 0f)
+                    };
+
+                    Pawn.TakeDamage(damageInfo);
+                }
```
mzegar, line 75: Should prediction be turned off for damage related things? Could this cause issues? This might be a case where we want to leave it on.

code/player/controller/DefaultWalkController.cs
```diff
@@ -43,9 +52,75 @@ public override void Simulate()
...
+                    DamageInfo damageInfo = new DamageInfo
+                    {
+                        Attacker = Pawn,
+                        Flags = DamageFlags.Fall,
+                        HitboxIndex = (int) HitboxIndex.LeftFoot,
+                        Position = Position,
+                        Damage = (MathF.Abs(_fallVelocity) - FallDamageVelocity) * FallDamageScale
+                    };
+
+                    Pawn.TakeDamage(damageInfo);
```
mzegar, line 119: Likewise here, another case where leaving prediction on might be the best option.

code/ui/DamageIndicator.cs
```diff
@@ -0,0 +1,69 @@
...
+        {
+            if (Host.IsServer)
+            {
+                return;
+            }
+
+            _lastDamage = damage;
+            _timeSinceLastDamage = 0f;
+            _additionalDamageIndicatorTime += _currentRemainingDamageIndicatorTime;
+            _currentRemainingDamageIndicatorTime = 0f;
```
mzegar, line 36: `Time` => `Duration` might just sound a bit better.

code/ui/DrowningIndicator.cs
```diff
@@ -0,0 +1,44 @@
...
+
+            _drowningBar = new BarPanel(this, "100", "drowningbar");
+
+            SetClass("hide", true);
+        }
+
+        public override void Tick()
+        {
+            if (Local.Pawn is not TTTPlayer player
+            || player.Controller is not DefaultWalkController defaultWalkController)
```
mzegar, line 26: Do we need this double line?

code/ui/DrowningIndicator.cs
```diff
@@ -0,0 +1,44 @@
...
+            {
+                return;
+            }
+
+            if (defaultWalkController.IsDiving)
+            {
+                float leftDivingTime = MathF.Max(defaultWalkController.DrownDamageTime - defaultWalkController.TimeSinceDivingStarted, 0f);
```
mzegar, line 33: `leftDivingTime` => `RemainingDivingDuration`, `TimeLeftDiving`, `RemainingBreath`, `BreathRemaining`

code/player/controller/DefaultWalkController.cs
Alf21: Controller should be predicted in any way. So stuff that is needed to calculate predicted stuff has to be the same on both sides.  I could make this var predicted, but as it's clientside just used for smooth visual effects and serverside used to adjust the health / dealing damage, I think this is the best way avoiding unnecessary network traffic.

code/player/controller/DefaultWalkController.cs
Alf21: As damage shouldn't be predictable (or am I wrong?) and shouldn't be send multiple times, I wouldn't turn it on here

mzegar (approved):
LGTM

PR #112, opened 2021-07-11
UI: Refactored folder structure

Alf21:
Based on #111

PR #113, opened 2021-07-11
UI: Added Spectator UI panel

Alf21:
image

Based on #112 and #111


Westsaid:
I really like that you've put the name of the spectated player in the left bottom corner. We might want to change the color, i'll try to look up a suiting color, something like a yellow/orange color.

Alf21:
> I really like that you've put the name of the spectated player in the left bottom corner. We might want to change the color, i'll try to look up a suiting color, something like a yellow/orange color.

Color is black for ROLE.None and the role color if you are aware of the player's role

code/ui/deadhud/SpectatedPlayerInfo.cs
```diff
@@ -0,0 +1,130 @@
...
+            public override void Tick()
+            {
+                base.Tick();
+
+                if (_currentPlayer == null)
+                {
+                    return;
+                }
+
+                if (_currentPlayer.Role is NoneRole)
```
mzegar, line 73: Can`NoneRole` just contain the colour?

code/ui/deadhud/SpectatedPlayerInfo.cs
```diff
@@ -0,0 +1,130 @@
...
+                    Style.BackgroundColor = _currentPlayer.Role.Color;
+                }
+
+                Style.Dirty();
+
+                _nameLabel.Text = $"{_currentPlayer.GetClientOwner()?.Name}";
+            }
+        }
+
+        private class IndicatorsPanel : Panel
```
mzegar, line 88: This is a copy and paste of how we do the health HUD on an alive player right? Is there any way to just reuse the code?

code/ui/deadhud/SpectatedPlayerInfo.cs
Alf21: It already has a color, but not the color we want (transparent).

code/ui/deadhud/SpectatedPlayerInfo.cs
Alf21: We could put it into a shared folder, but that's quite ugly. Additionally, we don't know what we gonna change in the future. One thing we could do is to use the same scss.

mzegar (approved):
LGTM

PR #115, opened 2021-07-16
UI: Added `VoiceChat` and `TeamVoiceChat`

Alf21:
unknown (4)

TODO:
- [x] Discuss how to display / separate team voice chat and global voice chat


mzegar:
@Alf21 I'm guessing we are going to keep this up meanwhile the bug report exists?

Alf21:
Hmmm... Currently I can't think of a way fixing it, it has to be done sbox-sided, so we don't need to wait for anything... Let me remove the error printing and adjust the bugreport to a commit link.

Could you test whether you can reproduce this issue (maybe it's caused by my setup),

code/player/TTTPlayer.VoiceChat.cs
```diff
@@ -0,0 +1,177 @@
...
+                            clients.Add(client);
+                        }
+                    }
+                }
+
+                _oldReceiveClients[player] = clients;
+            }
+            else
+            {
+                // player did not talked before
```
mzegar, line 92: Just minor grammar, `Player did not talk before`, or `Player has not talked before`

mzegar (approved):
Seems reasonable enough

PR #130, opened 2021-07-17
Team: Rework to work with class based Teams instead of strings

Alf21:


code/global/RPCs.cs
```diff
@@ -52,7 +52,14 @@ public static void ClientSetRole(TTTPlayer player, string roleName, string teamN
                 return;
             }
 
-            player.SetRole(Utils.GetObjectByType<TTTRole>(Utils.GetTypeByName<TTTRole>(roleName)), TTTTeam.GetTeam(teamName));
+            TTTTeam team = null;
+
+            if (teamName != null)
+            {
+                team = TeamFunctions.GetTeam(teamName);
```
mzegar, line 59: Can we just handle the `teamName` null check inside of `GetTeam()`?

code/teams/TraitorTeam.cs
```diff
@@ -0,0 +1,13 @@
+namespace TTTReborn.Teams
+{
+    [TeamAttribute("Traitors")]
+    public class TraitorTeam : TTTTeam
+    {
+        public override Color Color => Color.FromBytes(223, 41, 53);
```
mzegar, line 6: Can we just use regular R, G, B values here?

mzegar (commented):
Quick comments

code/teams/TraitorTeam.cs
Alf21: it is RGB. What do you mean? :)

mzegar (approved):
LGTM

PR #131, opened 2021-07-17
Prediction: Removed useless Prediction.Off in weapons

Alf21:


code/items/weapons/Knife.cs
```diff
@@ -57,10 +57,7 @@ public override void AttackPrimary()
         {
             (Owner as AnimEntity).SetAnimBool("b_attack", true);
 
-            using (Prediction.Off())
-            {
-                ShootEffects();
-            }
+            ShootEffects();
```
mzegar, line 60: Good catch, seems like we shouldn't be turning off prediction for this.

https://github.com/Facepunch/sbox-hidden/blob/dbe2903ec3af243715a34275414f63fa49b88400/code/weapons/SMG.cs#L36

mzegar (approved):
LGTM

code/items/weapons/Knife.cs
Alf21: Maybe this helps in some way. Anyways, doesn't explain the other issues we had with hearing yourself talking, animation issue etc. (this is just for the shooting effects).

PR #132, opened 2021-07-17
UI: Inventory and Chat rework

Alf21:
image
image

- Chat messages of the same player are grouped now
- ChatMessages won't be removed anymore (just if the limit is reached)
- Chat is scrollable now
- Chat will try to autoscroll to the end if you're not already scrolling
- Perks were placed next to the inventory


AMastryukov (approved):

Westsaid:
Good work finishing and tweaking this

PR #133, opened 2021-07-17
Feature/inventory system

AMastryukov:
- Add primary, secondary, melee, equipment and grenade slots
- Change relevant areas to use SlotType instead of HoldType
- Replace hardcoded HoldType with variable values
- Add inventory slot capacity logic
- Add random SMG/Shotgun on spawn (either one 50-50 chance)
- Remove redundant code from weapon implementations
- Adjust reload times to match animations better (when magazine is inserted)
- Update shop UI to reflect new slot behavior
- Update weapon selection UI for unlimited ammo weapons

### DIFF
--- a/code/items/IBuyableItem.cs
+++ b/code/items/IBuyableItem.cs
@@ -1,4 +1,5 @@
 using System;
+
 using Sandbox;
 
 using TTTReborn.Player;
--- a/code/items/IBuyableItem.cs
+++ b/code/items/IBuyableItem.cs
@@ -1,4 +1,5 @@
 using System;
+using Sandbox;
 
 using TTTReborn.Player;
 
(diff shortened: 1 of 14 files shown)


code/player/Inventory.cs
```diff
@@ -10,8 +10,8 @@ namespace TTTReborn.Player
     public partial class Inventory : BaseInventory
     {
         public readonly PerksInventory Perks;
-
         public readonly AmmoInventory Ammo;
+        public int[] SlotCapacity => new int[] { 1, 1, 1, 3, 3 };
```
Alf21, line 14:
```suggestion
        public readonly int[] SlotCapacity => new int[] { 1, 1, 1, 3, 3 };
```

code/player/Inventory.cs
```diff
@@ -42,13 +42,13 @@ public override bool Add(Entity entity, bool makeActive = false)
         {
             TTTPlayer player = Owner as TTTPlayer;
 
-            if (IsCarryingType(entity.GetType()))
-            {
-                return false;
-            }
```
Alf21, line 48: Shouldn't this be kept if you trying to pickup something you already have that is not derived from `ICarriableItem`?

code/ui/alivehud/inventoryselection/InventorySelection.cs
```diff
@@ -186,6 +186,11 @@ private int GetKeyboardNumberPressed(InputBuilder input)
 
         private static string FormatAmmo(TTTWeapon weapon, Inventory inventory)
         {
+            if (weapon.UnlimitedAmmo)
+            {
+                return $"{weapon.AmmoClip}";
+            }
```
Alf21, line 192: Good catch!

code/items/IBuyableItem.cs
```diff
@@ -9,21 +11,20 @@ public struct ShopItemData
         public string Name;
         public string Description;
         public int Price;
+        public SlotType SlotType;
```
Alf21, line 14: That's the only change I don't like in this PR, as Perks are `IBuyableItems` as well, but never have a `SlotType`

mzegar:
Don't have anything to add, looks fine to me, check out Alf comments :D

code/player/Inventory.cs
AMastryukov: Damn, good catch. I will revert.

code/items/IBuyableItem.cs
AMastryukov: I agree, I don't like this change either.

We should discuss what alternatives exist to this approach of checking if players are able to purchase a weapon/equipment when there is no slot available.

code/player/Inventory.cs
AMastryukov: Reverted

code/items/IBuyableItem.cs
Alf21: `SlotType` still could be a part of `ICarriableItem`, then you just need to check for it explicitly.

Alf21 (approved):
Nice improvements :)

Issue #118, opened 2021-07-17
Damage Icons


Alf21:
[ ] - Drowning
[ ] - Fall Damage
[ ] - Headshot

Westsaid:
I got a friend who made the icons which are used on http://www.ttt-sbox.com/, we can use some of them and also ask him to make other fresh icons

PR #134, opened 2021-07-18
InspectMenu: Added data

Alf21:
image

Added to InspectMenu:
- KillerWeapon
- Time the player was killed
- Headshot indicator
- Distance indicator


Alf21:
@TimGoll Maybe you could add some more matching Icons? Regarding `Headshot`, `TimeSinceDeath` and #118

code/player/TTTPlayer.Damage.cs
```diff
@@ -47,29 +46,45 @@ public partial class TTTPlayer
...
                 info.Damage *= 2.0f;
             }
 
+            LastDamageWeapon = info.Weapon.IsValid() ? info.Weapon as TTTWeapon : null;
+
             To client = To.Single(this);
 
             if (info.Attacker is TTTPlayer attacker && attacker != this)
             {
+                LastDistanceToAttacker = Utils.SourceUnitsToMeters(Position.Distance(attacker.Position));
```
AMastryukov, line 75: I converted this meaningless Source distance value to meters

AMastryukov (dismissed):

code/player/TTTPlayer.Confirmation.cs
```diff
@@ -1,9 +1,20 @@
+using System.Collections.Generic;
+
 using Sandbox;
 
 using TTTReborn.Globals;
 
 namespace TTTReborn.Player
 {
+    public struct ConfirmationData
```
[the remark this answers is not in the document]
Alf21: Try to add strings to networked structs :>

mzegar:
Do we want to remove the prefix so the text stays on one line?

`00:12s`, `81m`

AMastryukov:
> Do we want to remove the prefix so the text stays on one line?
> 
> `00:12s`, `81m`

Not sure what you mean by this. You want to remove the extra text and just keep the number format and that's it?
If so, I'm neither for nor against it. I like the way it currently is, but I also like it simpler.

One thing that's cool about this way is that it kind of reads like a sentence: "Died 00:01 ago by headshot from 10m away with a shotgun"

mzegar:
> > Do we want to remove the prefix so the text stays on one line?
> > `00:12s`, `81m`
> 
> Not sure what you mean by this. You want to remove the extra text and just keep the number format and that's it?
> If so, I'm neither for nor against it. I like the way it currently is, but I also like it simpler.
> 
> One thing that's cool about this way is that it kind of reads like a sentence: "Died 00:01 ago by headshot from 10m away with a shotgun"

Yeah, whatever removes the double lines imo.

Alf21:
> > > Do we want to remove the prefix so the text stays on one line?
> > > `00:12s`, `81m`
> > 
> > 
> > Not sure what you mean by this. You want to remove the extra text and just keep the number format and that's it?
> > If so, I'm neither for nor against it. I like the way it currently is, but I also like it simpler.
> > One thing that's cool about this way is that it kind of reads like a sentence: "Died 00:01 ago by headshot from 10m away with a shotgun"
> 
> Yeah, whatever removes the double lines imo.

Yea I thought we should display the minimum information below the icon "00:06s" and when hovering in detailed mode, display "Died 00:06s ago" e.g.

mzegar:
> > > > Do we want to remove the prefix so the text stays on one line?
> > > > `00:12s`, `81m`
> > > 
> > > 
> > > Not sure what you mean by this. You want to remove the extra text and just keep the number format and that's it?
> > > If so, I'm neither for nor against it. I like the way it currently is, but I also like it simpler.
> > > One thing that's cool about this way is that it kind of reads like a sentence: "Died 00:01 ago by headshot from 10m away with a shotgun"
> > 
> > 
> > Yeah, whatever removes the double lines imo.
> 
> Yea I thought we should display the minimum information below the icon "00:06s" and when hovering in detailed mode, display "Died 00:06s ago" e.g.

Seems reasonable to me

AMastryukov:
> > > > Do we want to remove the prefix so the text stays on one line?
> > > > `00:12s`, `81m`
> > > 
> > > 
> > > Not sure what you mean by this. You want to remove the extra text and just keep the number format and that's it?
> > > If so, I'm neither for nor against it. I like the way it currently is, but I also like it simpler.
> > > One thing that's cool about this way is that it kind of reads like a sentence: "Died 00:01 ago by headshot from 10m away with a shotgun"
> > 
> > 
> > Yeah, whatever removes the double lines imo.
> 
> Yea I thought we should display the minimum information below the icon "00:06s" and when hovering in detailed mode, display "Died 00:06s ago" e.g.

Even with these changes, since other stuff is dynamic it may bleed into new lines. For example, the weapon names.

mzegar:
Does s&box support text overflow options? I would personally just force a `ellipsis` on any extremely long name. From what it seems like it's only weapon names that we would need to be concerned with.

I'm open to hear opinions.

AMastryukov:
New weapon icons:
image

AMastryukov (approved):

code/player/TTTPlayer.Confirmation.cs
```diff
@@ -114,6 +136,21 @@ private void BecomePlayerCorpseOnServer(Vector3 force, int forceBone)
                 Rotation = Rotation
             };
 
+            corpse.KillerWeapon = LastDamageWeapon?.Name;
+            corpse.WasHeadshot = LastDamageWasHeadshot;
+            corpse.Distance = LastDistanceToAttacker;
+            corpse.Suicide = LastAttacker == this;
+
+            List<string> perks = new();
```
mzegar, line 144: Why not just make this an array?

code/player/TTTPlayer.Confirmation.cs
```diff
@@ -114,6 +136,21 @@ private void BecomePlayerCorpseOnServer(Vector3 force, int forceBone)
...
+            PerksInventory perksInventory = (Inventory as Inventory).Perks;
+
+            for (int i = 0; i < perksInventory.Count(); i++)
+            {
+                perks.Add(perksInventory.Get(i).Name);
+            }
+
+            corpse.Perks = perks.ToArray();
```
mzegar, line 152: This logic could be cleaned up.

code/ui/alivehud/inspectmenu/InspectMenu.cs
```diff
@@ -39,7 +40,7 @@ public InspectMenu()
             _confirmationPanel = new ConfirmationPanel(this);
         }
 
-        public void InspectCorpse(TTTPlayer deadPlayer, bool isIdentified)
+        public void InspectCorpse(TTTPlayer deadPlayer, bool isIdentified, ConfirmationData confirmationData, string killerWeapon = null, string[] perks = null)
```
mzegar, line 43: Is there a reason why we can't package `KillerWeapon` and `Perks` inside of `confirmationData`? As well as `isIdentified`?

code/ui/alivehud/inspectmenu/InspectMenu.cs
```diff
@@ -132,6 +151,88 @@ public void SetPlayer(TTTPlayer player)
...
+                        _distance = new InspectItem(this);
+                        _distance.ImageWrapper.Image.SetTexture($"/ui/inspectmenu/distance.png");
+                        _distance.InspectItemLabel.Text = $"From {confirmationData.Distance:n0}m away";
+                    }
+                    else
+                    {
+                        _suicide = new InspectItem(this);
+                        _suicide.ImageWrapper.Image.SetTexture("");
+                        _suicide.InspectItemLabel.Text = $"Committed suicide";
+                    }
```
mzegar, line 188: Can we not just update the data instead of deleting and re-creating these panels?

code/ui/alivehud/inspectmenu/InspectMenu.cs
Alf21: https://github.com/TTTReborn/tttreborn/pull/134#discussion_r671869036

code/ui/alivehud/inspectmenu/InspectMenu.cs
Alf21: We could but things not always exist for every confirmation stuff and then it could get reordered. That here makes sure that it's always in the normal order.

code/ui/alivehud/inspectmenu/InspectMenu.cs
mzegar: So we could pass in `bool isIdentified`

code/player/TTTPlayer.Confirmation.cs
Alf21: Good catch, feel free changing it as I don't get time doing so until tomorrow.

code/ui/alivehud/inspectmenu/InspectMenu.cs
mzegar: I'm not sure I understand, if the components are created once, and then updated, how could they end up being re-ordered?

Alf21: As there gonna be a lot of ui icon elements, hiding is not an option. So we have to delete them. And if you e.g. delete the headshot entry, it will - if there is an headshot on another corpse - be created and *appended* to the end of the element (instead of being placed in the mid / as second or third element e.g.).

PR #138, opened 2021-07-18
Spectator: Added first person spectator camera

Alf21:


mzegar (changes requested):
1. We should remove any transition time between swapping players
2. You can also spectate other spectators as well, this should be fixed.

Alf21:
> 1. Is there a way to remove transition time between players?
> 2. You can also spectate other spectators as well, this should be fixed.

1. Wdym?
2. I used the same logic of the other spec cams. If this is a bug there, it should be a bug in all spec cams.

Don't have the time working on it in the next days, so feel free modifying it on your own :)

mzegar:
> > 1. Is there a way to remove transition time between players?
> > 2. You can also spectate other spectators as well, this should be fixed.
> 
> 1. Wdym?
> 2. I used the same logic of the other spec cams. If this is a bug there, it should be a bug in all spec cams.
> 
> Don't have the time working on it in the next days, so feel free modifying it on your own :)

1. When you swap spectating views, I think the camera "travels" as opposed to teleports to the next player. I haven't had any luck with fixing this.
2. Yeah after the spectating merge we should fix that spectating issue, I'll see if I can take a look sometime tonight maybe.

Alf21:
> 1. When you swap spectating views, I think the camera "travels" as opposed to teleports to the next player. I haven't had any luck with fixing this.

This should be fixed now

code/player/TTTPlayer.Lookup.cs
```diff
@@ -8,20 +8,32 @@ partial class TTTPlayer
...
+                return type;
             }
 
-            return null;
+            return default(T);
+        }
+
+        private TTTPlayer IsLookingAtPlayer()
+        {
+            return IsLookingAtType<TTTPlayer>(MAX_DRAW_DISTANCE);
```
mzegar, line 36: Do we want to make a IsLookingAtPlayerCorpse? Don't really mind if not. The interface cleans it up a bunch in general.

code/player/TTTPlayer.Spectating.cs
```diff
@@ -0,0 +1,52 @@
...
+                return;
+            }
+
+            if (player.IsSpectatingPlayer && player.CurrentPlayer == deadPlayer)
+            {
+                player.UpdateObservatedPlayer();
+            }
+        }
+
+        public void UpdateObservatedPlayer()
```
mzegar, line 28: Do we want to instead go with `Spectated` instead of `Observated`?

code/player/camera/FirstPersonSpectatorCamera.cs
```diff
@@ -0,0 +1,67 @@
...
+            else
+            {
+                Pos = Vector3.Lerp(_lastPos, player.CurrentPlayer.EyePos, SMOOTH_SPEED * Time.Delta);
+                Rot = Rotation.Slerp(_lastRot, player.CurrentPlayer.EyeRot, SMOOTH_SPEED * Time.Delta);
+            }
+
+            _lastPos = Pos;
+            _lastRot = Rot;
+
+            FieldOfView = 80;
```
mzegar, line 42: I think you can just set this up once on `Activated()` method. Instead of setting it each time on `Update()`. I hope that our other cameras don't do this as well.

code/player/camera/FirstPersonSpectatorCamera.cs
```diff
@@ -0,0 +1,67 @@
...
+        }
+
+        public void OnUpdateObservatedPlayer(TTTPlayer oldObservatedPlayer, TTTPlayer newObservatedPlayer)
```
mzegar, line 45: Same here, if we decide that we should go with `Spectated` instead of `Observated`

code/player/camera/ThirdPersonSpectateCamera.cs
```diff
@@ -36,17 +37,7 @@ public override void Update()
...
-                if (players.Count > 0)
-                {
-                    if (++_targetIdx >= players.Count)
-                    {
-                        _targetIdx = 0;
-                    }
-
-                    player.CurrentPlayer = players[_targetIdx];
-                }
+                player.UpdateObservatedPlayer();
```
mzegar, line 40: Should we have a param into `UpdateObservatedPlayer` that either decrements or increments? Goes forwards and backwards? Or we can leave this in for another PR as well.

mzegar (commented):
Comments

code/player/TTTPlayer.Spectating.cs
mzegar: I guess technically it could be specific to `Observation`, up to you, both are fine with me.

code/player/TTTPlayer.Lookup.cs
Alf21: Wouldn't do so as the corresponding distance is saved in another file :o

code/player/TTTPlayer.Spectating.cs
Alf21: Maybe in another PR to change every `observate` into `spectate`. I think I've choosen `observate` as there have been some issues using `spectate` I can't remember currently.

mzegar: np, I'm fine with both.

code/player/camera/ThirdPersonSpectateCamera.cs
Alf21: I'd just add this if it's needed at some point or requested :)

mzegar (approved):
Biggest thing would be camera thing, rest is optional. LGTM, nice work again.

PR #139, opened 2021-07-20
Code Refactoring: Simplified structure and calls

Alf21:


code/ui/alivehud/inventorywrapper/inventoryselection/InventorySelection.cs
```diff
@@ -175,28 +175,43 @@ private int NormalizeSlotIndex(int index, int maxIndex)
 
         private int GetKeyboardNumberPressed(InputBuilder input)
         {
-            if (input.Pressed(InputButton.Slot1)) return 1;
-            if (input.Pressed(InputButton.Slot2)) return 2;
-            if (input.Pressed(InputButton.Slot3)) return 3;
-            if (input.Pressed(InputButton.Slot4)) return 4;
-            if (input.Pressed(InputButton.Slot5)) return 5;
+            InputButton[] pressedButton = new[] {
```
AMastryukov, line 178: This should be declared outside of the method

code/ui/alivehud/inventorywrapper/inventoryselection/InventorySelection.cs
```diff
@@ -175,28 +175,43 @@ private int NormalizeSlotIndex(int index, int maxIndex)
...
-            return 0;
+            return -1;
         }
 
         private static string FormatAmmo(TTTWeapon weapon, Inventory inventory)
         {
             if (weapon.UnlimitedAmmo)
             {
-                return $"{weapon.AmmoClip}";
+                return $"{weapon.AmmoClip} + ∞";
```
AMastryukov, line 206: I tried this before and it didn't look very good.
I prefer having no "+ ∞" after as a single value should be enough to convey that the reserve is infinite.

AMastryukov:
I like most of the changes, but I really don't like the changes where the new `ObjectType()` is changed to just `new()`, especially when the types are not immediately obvious. 

I think it's fine in cases like this:
`public KarmaSystem Karma { get; private set; } = new()` 

but not in cases like this:
`_healthBar = new(this, "", "healthbar")`

I immediately think that `_healthBar` is of type `HealthBar`, when in reality it's actually a `BarPanel`.

AMastryukov (changes requested):
I like most of the changes, but I really don't like the changes where the new `ObjectType()` is changed to just `new()`, especially when the types are not immediately obvious. 

I think it's fine in cases like this:
`public KarmaSystem Karma { get; private set; } = new()` 

but not in cases like this:
`_healthBar = new(this, "", "healthbar")`

I immediately think that `_healthBar` is of type `HealthBar`, when in reality it's actually a `BarPanel`.

code/ui/alivehud/inventorywrapper/inventoryselection/InventorySelection.cs
```diff
@@ -175,28 +188,30 @@ private int NormalizeSlotIndex(int index, int maxIndex)
...
-            if (input.Pressed(InputButton.Slot3)) return 3;
-            if (input.Pressed(InputButton.Slot4)) return 4;
-            if (input.Pressed(InputButton.Slot5)) return 5;
+            for (int i = 0; i < _slotInputButtons.Length; i++)
+            {
+                if (input.Pressed(_slotInputButtons[i]))
+                {
+                    return i;
+                }
+            }
```
mzegar, line 197: Could just create a map of `InputButton` -> `i` and then try and fetch the value from the map.

code/ui/alivehud/inventorywrapper/inventoryselection/InventorySelection.cs
mzegar: Hmm, no opinion on my end, we probably will just remove infinite ammo at some point anyways. Flip a coin or ask some other people to chime in as a tie breaker.

Alf21:
> I like most of the changes, but I really don't like the changes where the new `ObjectType()` is changed to just `new()`, especially when the types are not immediately obvious.
> 
> I think it's fine in cases like this:
> `public KarmaSystem Karma { get; private set; } = new()`
> 
> but not in cases like this:
> `_healthBar = new(this, "", "healthbar")`
> 
> I immediately think that `_healthBar` is of type `HealthBar`, when in reality it's actually a `BarPanel`.

On the other side, if you e.g. see `list.Add()`, you don't know whether it's a `List`, `Map`, `Dictionary` etc. as well. I like using `new` to make sure you don' try to set it to another, maybe inherited type.

AMastryukov:
> > I like most of the changes, but I really don't like the changes where the new `ObjectType()` is changed to just `new()`, especially when the types are not immediately obvious.
> > I think it's fine in cases like this:
> > `public KarmaSystem Karma { get; private set; } = new()`
> > but not in cases like this:
> > `_healthBar = new(this, "", "healthbar")`
> > I immediately think that `_healthBar` is of type `HealthBar`, when in reality it's actually a `BarPanel`.
> 
> On the other side, if you e.g. see `list.Add()`, you don't know whether it's a `List`, `Map`, `Dictionary` etc. as well. I like using `new` to make sure you don' try to set it to another, maybe inherited type.

Right, which is why naming is so important. I wouldn't name a Dictionary a `list`, I'd name it `dictionary`.
I dunno, I'm still uneasy with this change, but if others disagree with me then I'll back down.

Alf21:
> > > I like most of the changes, but I really don't like the changes where the new `ObjectType()` is changed to just `new()`, especially when the types are not immediately obvious.
> > > I think it's fine in cases like this:
> > > `public KarmaSystem Karma { get; private set; } = new()`
> > > but not in cases like this:
> > > `_healthBar = new(this, "", "healthbar")`
> > > I immediately think that `_healthBar` is of type `HealthBar`, when in reality it's actually a `BarPanel`.
> > 
> > 
> > On the other side, if you e.g. see `list.Add()`, you don't know whether it's a `List`, `Map`, `Dictionary` etc. as well. I like using `new` to make sure you don' try to set it to another, maybe inherited type.
> 
> Right, which is why naming is so important. I wouldn't name a Dictionary a `list`, I'd name it `dictionary`.
> I dunno, I'm still uneasy with this change, but if others disagree with me then I'll back down.

Yea same for me. Just suggestion on my side, I don't have a problem if we decide against that single change.

So let's vote:

👍 = `var = new()`
👎 = `var = new VarType()`

Anyways, I think we all agree using `VarType var = new()` anyways.

mzegar:
@Alf21 is there anything else we need here? You can ignore my dictionary suggestion, won't work.

mzegar (approved):
LGTM, I know it's draft, but it seems to be mostly just IDE simplifications.

Alf21:
Just to decide the `endless ammo` change and to merge the PR :p

mzegar (changes requested):
Merge conflicts need to be handled

AMastryukov (approved):

mzegar (approved):

PR #140, opened 2021-07-20
UI: Support both alive & spectating players in UI

Alf21:
This should reduce code redundancy and make it easier to created observing-support for ui elements.

Inventory, PlayerInfo, Nameplate and InspectMenu will be shown for spectators now as well

image


mzegar:
Right yeah, this allows us to reuse UI elements. I'm going to see if there is any other way to implement this differently (there might not be), but I like this idea.

mzegar:
https://github.com/TTTReborn/tttreborn/pull/141

Does this work? Then we can just reference `ObservablePlayer` in any places in the HUD where we need to.

Alf21:
> #141
> 
> I came up with this, does this seem fine? Then we can just reference `ObservablePlayer` in any places in the HUD where we need to.

I like these changes, but you have to reset this var properly e.g. on cam switch and you should integrate it into this PR directly :)

mzegar:
> > #141
> > I came up with this, does this seem fine? Then we can just reference `ObservablePlayer` in any places in the HUD where we need to.
> 
> I like these changes, but you have to reset this var properly e.g. on cam switch and you should integrate it into this PR directly :)

Sounds good, yeah I avoided adding that in as it was just to showcase the idea. I'll see if I can get this done sometime this weekend.

mzegar:
There were a bit too many changes on this branch, just wanted to start fresh. It should include most of the changes you had on this branch.

https://github.com/TTTReborn/tttreborn/pull/142

Closing this PR, the branch will still exist in case you want to bring it back.

Alf21:
> There were a bit too many changes on this branch, just wanted to start fresh. It should include most of the changes you had on this branch.
> 
> #142
> 
> Closing this PR, the branch will still exist in case you want to bring it back.

Sorry, but why? I mean those changes are in the other PR as well ^^

mzegar:
Sure I can force push changes to this branch if you'd like.

mzegar (approved):
@Alf21 ready whenever you want to merge.

PR #144, opened 2021-07-24
Scoreboard: Rework and MIA/Dead re-positioning fix

Alf21:


code/ui/generalhud/scoreboard/Scoreboard.ScoreboardGroup.cs
```diff
@@ -0,0 +1,118 @@
...
+                            group = DefaultScoreboardGroup.Dead.ToString();
+                        }
+                        else if (player.IsMissingInAction)
+                        {
+                            group = DefaultScoreboardGroup.MIA.ToString();
+                        }
+
+                        break;
+                    }
+                }
```
mzegar, line 101: Might be able to use our "GetAllPlayers" util method here.

mzegar (commented):
Minor comment

code/ui/generalhud/scoreboard/Scoreboard.ScoreboardGroup.cs
Alf21: Uhm do you have a link to that method? I don't find anyone that would make sense to be integrated here sadly :/

mzegar: nvm, doesn't make sense in this case.

mzegar (approved):
Seems reasonable to me? I like that we threw the header into a new file. It was getting a bit long.

PR #150, opened 2021-07-25
VoiceChat: Added key binding support to TeamVoiceChat

Alf21:
To use, you have to add a binding like this:

`bind b +teamvoicechat`


mzegar (approved):
LGTM, I wonder where we should keep this information, but probably something to consider in the future. (Like how do we let players know about this bind)

PR #154, opened 2021-07-25
Fix: Postround sync

Alf21:
- Fixed dead players do not refresh on round end
- Fixed dead players do not get confirmed on round end
- Fixed InspectMenu is getting overwritten with different PlayerCorpse data if looking at a PlayerCorpse when another PlayerCorpse is confirmed
- Fixed Scoreboard update issue
- Removed unused score "alive" state


mzegar (approved):
Nice fixes

PR #155, opened 2021-07-26
Bug: Fix camera changing when user connects midround

mzegar:


mzegar:
We should handle this using lifestate.

PR #158, opened 2021-07-26
Added chat cmd support (!...)

Alf21:


mzegar:
Was this popular in GMOD? I'd prefer to keep commands in console.

Alf21:
Was a thing as you would be able to modify / add autocomplete and useful tools like running commands multiple times etc

mzegar (approved):
LGTM

Alf21:
Waiting with this PR for https://github.com/Facepunch/sbox-issues/issues/501

code/ui/generalhud/chat/ChatBox.cs
```diff
@@ -111,11 +114,38 @@ private void Submit()
...
+                {
+                    if (!cr.WasFound)
+                    {
+                        AddEntry("SYSTEM", $"Cmd: '{cmd}' was not found!");
+
+                        return;
+                    }
+
+                    AddEntry("SYSTEM", $"Error with cmd: '{cmd}'!");
+                }
```
mzegar, line 144: Might be worth to throw this stuff into a function.

mzegar:
Did Garry end up fixing this? @Alf21 https://github.com/Facepunch/sbox-issues/issues/501

Alf21:
Nope, he just removed `ConsoleResult` and closed this as this won't ever get supported. Instead, maybe can be used with reflections or a pre defined admin cmd list...

PR #165, opened 2021-07-29
UI: Keybind Menu

Alf21:
Still WIP as `ConsoleResult.Value` is `null` as well and there is no other way to access the console output or currently bound keys. Waiting for https://github.com/Facepunch/sbox-issues/issues/501 or https://github.com/Facepunch/sbox-issues/issues/172

### DIFF
--- a/code/ui/generalhud/menu/Menu.cs
+++ b/code/ui/generalhud/menu/Menu.cs
@@ -1,0 +1,95 @@
+using Sandbox.UI;
+using Sandbox.UI.Construct;
+
+namespace TTTReborn.UI
+{
+    public class Menu : Panel
+    {
+        public static Menu Instance;
+
+        public bool IsShowing
+        {
+            get => _isShowing;
+            set
+            {
+                _isShowing = value;
+
+                SetClass("hide", !_isShowing);
+            }
+        }
+
+        private bool _isShowing = false;
+
+        private KeybindEntry _keybindEntry;
+
+        public Menu()
+        {
+            Instance = this;
+
+            StyleSheet.Load("/ui/generalhud/menu/Menu.scss");
+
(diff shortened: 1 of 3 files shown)


code/ui/generalhud/menu/Menu.cs
```diff
@@ -0,0 +1,95 @@
+using Sandbox.UI;
+using Sandbox.UI.Construct;
+
+namespace TTTReborn.UI
+{
+    public class Menu : Panel
```
mzegar, line 6: Do we want to make this name more specific, is this basically an options menu?

code/ui/generalhud/menu/Menu.cs
```diff
@@ -0,0 +1,95 @@
...
+            get => _isShowing;
+            set
+            {
+                _isShowing = value;
+
+                SetClass("hide", !_isShowing);
+            }
+        }
+
+        private bool _isShowing = false;
```
mzegar, line 21: I wonder if we should have our own custom derived class that has this functionality, and each UI element can inherit from it, what do you think? We can add this as an issue.

code/ui/generalhud/menu/Menu.cs
```diff
@@ -0,0 +1,95 @@
...
+
+        public Menu()
+        {
+            Instance = this;
+
+            StyleSheet.Load("/ui/generalhud/menu/Menu.scss");
+
+            _keybindEntry = Add.KeybindEntry("Press a button");
+
+            SetClass("hide", true);
```
mzegar, line 33: Could technically use the property you made.

Alf21:
Moved into https://github.com/TTTReborn/tttreborn/pull/177

Issue #63, opened 2021-06-25
Ammo System


AMastryukov:
Ammo should be stored based on type, and new types should be easily added if required (e.g. a special purchasable weapon that uses a type of ammo that the players cannot acquire in any other way and that is not shared with any other weapon).

Fires1:
Going to toss in my suggested change to this, you can find my attempted solution at Fires1/tttreborn@e330250 .

So with this change the ammo system moves from a fixed AmmoType enum to a growing dictionary of string keys for each specific ammo type.
The ammo type is registered within a players inventory during [TTTWeapon.Spawn](https://github.com/Fires1/tttreborn/blob/e3302509e15b348fcc16770c1cb303a34e658576/code/items/weapons/TTTWeapon.cs#L109). So the player's inventory should only hold values for weapons they have picked up. That is atleast my understanding of how the process currently works within weapon spawning in S&box. 
I haven't encountered any issues within my testing, weapons with the same ammo type have a shared pool of ammo as expected, and new ammo types can be registered and utilized by simply overriding the AmmoType *string* within a weapon definition.
Should be a somewhat cleaner solution versus hardcoding ammo types in.

This is my first contribution to this project, as well as my first bit of tinkering within S&box's environment so I'm still working on understanding fully and getting used to how everything operates. So just keep that in mind. Let me know how everything looks and thoughts on the overall system change.

mzegar:
@Fires1 do you want to open a PR for this?

Fires1:
Notes for some extension stuff to do after merging #167 .

- TTTAmmo.cs Line 57: Possible to optimize the ammo typing process by caching weapons within the ammo entity, then checking that list against the players, rather than a larger, complicated system.
- TTTWeapon.cs Line 164: Move input to drop ammo to a bindable key once that system is up and running.
- Sounds/animations on pickup.

Will update this comment with links as soon as merge is completed.

PR #167, opened 2021-07-31
Rework ammo inventory

Fires1:
[More information available within the issue comments here](https://github.com/TTTReborn/tttreborn/issues/63#issuecomment-890295627)


Fires1:
Best used as a expandable placeholder for the moment. Alf expressed last night about possibly growing this into an object oriented system for implementation of different ammo types. Thought about it more, it may be required in order to hold onto data relevant to a specific ammo type (such as the entity that is created when you drop ammo). 
Garry's Mod TTT used SWEP.AmmoEnt so it's possible we can just add an additional field to allow guns to drop/pickup ammo.

code/items/weapons/TTTWeapon.cs
```diff
@@ -104,6 +104,10 @@ public override void Spawn()
             base.Spawn();
 
             AmmoClip = ClipSize;
+            if (Owner is TTTPlayer owner)
```
Alf21, line 107:
```suggestion

            if (Owner is TTTPlayer owner)
```

code/player/Inventory.Ammo.cs
```diff
@@ -9,66 +9,92 @@ namespace TTTReborn.Player
 {
     public partial class AmmoInventory
     {
-        private List<int> AmmoList { get; set; } = new();
+        private Dictionary<string, int> AmmoList { get; set; } = new();
```
Alf21, line 12:
```suggestion
        private Dictionary<string, int> AmmoList { get; private set; } = new();
```

code/player/Inventory.Ammo.cs
```diff
@@ -9,66 +9,92 @@ namespace TTTReborn.Player
...
+            }
+
+            if (Host.IsServer)
+            {
+                TTTPlayer player = Inventory.Owner as TTTPlayer;
 
-            if (AmmoList == null || AmmoList.Count <= iAmmoType)
+                player.ClientRegisterAmmo(To.Single(player), ammo, amount);
+            }
+            return true;
```
Alf21, line 38:
```suggestion

            return true;
```

code/items/weapons/TTTWeapon.cs
```diff
@@ -104,6 +104,10 @@ public override void Spawn()
...
+            {
+                (owner.Inventory as Inventory).Ammo.RegisterType(AmmoType);
```
Alf21, line 109: Do we need to run a RPC to register stuff? Can't this be done in the SetAmmo RPC (and if no ammo was registered, returning 0 ammo)?

Alf21 (commented):
Thanks for your contribution ❤️ 
We talked about it and imo, I'd be okay approving this PR, but as already mentioned, this just gonna be a temporary solution (regarding object based ammo).

code/items/weapons/TTTWeapon.cs
```diff
@@ -161,6 +161,24 @@ public override void Simulate(Client owner)
...
             }
 
+            //No idea where else to put this. So it goes here.
+            if (Input.Released(InputButton.View) && AmmoClip > 0) //Todo: Move this to be a bindable key (default Z???)
+            {
+                if (IsServer)
+                {
+                    var ammoBox = new BuckshotAmmo()
+                    {
+                        AmmoAmount = AmmoClip,
```
Fires1, line 171: `AmmoAmount = AmmoClip` doesn't seem to actually work, thats why there is a call to `.SetCurrentAmmo` on line 176, which changes the private variable and sets it appropriately.

I think this is likely due to my minimal understanding of how the fuck S&box operates it's client-server communication. Would definitely like solutions or changes in cleaning up the networking logic as I get used to it myself.

Fires1:
Going to add in and adjust this PR to also include the ammo dropping stuff. I did **not** commit my stand-in placeholder model, but if need be I can (or create more, equally shitty, stand in models, for the other types)

As it stands, the dropping system works at least with one individual, but should be tested with other people to verify results are consistent.

Could use:
- [ ] Pickup sounds
- [ ] Some collision adjustments to make it so players cannot "climb" with ammo boxes
- [ ] Perhaps an animation when dropping?
- [ ] A notification if attempting to drop with no ammo?

Uses some finicky stuff to work around some issues (such as the networking with entity ammo counts). Am also unhappy with how the implementation of GetAmmoTypes works. I like the way it's implemented in the original TTT, where the ammo entity itself caches a list of weapons it's compatible with, then when making contact, it compares to. Cleaner and faster but unfortunately I could not find a S&box equivalent to Garry's Mod's `weapons.GetList()`

Definitely needs a lot of cleanup.

code/items/weapons/TTTWeapon.cs
Fires1: To clarify. Variable change on 171 doesn't work. Function call on 176 does.

code/items/weapons/TTTWeapon.cs
Fires1: You are correct. RPC removed, logic already exists within SetAmmo.

code/items/ammo/TTTAmmo.cs
```diff
@@ -0,0 +1,92 @@
...
+        }
+
+        public override void Touch(Entity other)
+        {
+            base.Touch(other);
+            if (IsServer)
+            {
+                if(other is TTTPlayer player)
+                {
+                    var ammoType = AmmoType.ToLower();
```
mzegar, line 49: Is this `ToLower()` call necessary?

code/items/ammo/BuckshotAmmo.cs
```diff
@@ -0,0 +1,13 @@
+using Sandbox;
+
+namespace TTTReborn.Items
+{
+    [Library("ttt_ammo_buckshot")]
+    partial class BuckshotAmmo : TTTAmmo
+    {
+        public override string AmmoType => "buckshot";
```
mzegar, line 8: Can this be an enum now? I might have missed some of the convo on the `string` discussion.

code/items/ammo/TTTAmmo.cs
```diff
@@ -0,0 +1,92 @@
...
+
+namespace TTTReborn.Items
+{
+    [Library("ttt_ammo"), Hammer.Skip]
+    public abstract partial class TTTAmmo : Prop
+    {
+        /// <summary>
+        /// String definition of ammo type, should match TTTWeapon.AmmoType
+        /// </summary>
+        public virtual string AmmoType { get; set; }
```
mzegar, line 13: If we are going to stick with `string`, I feel like `AmmoName` is a better name.

code/items/ammo/TTTAmmo.cs
```diff
@@ -0,0 +1,92 @@
...
+        /// <summary>
+        /// Amount of Ammo within Entity.
+        /// </summary>
+        public virtual int AmmoAmount { get; set; }
+
+        /// <summary>
+        /// Maximum amount of ammo player can carry in their inventory.
+        /// </summary>
+        public virtual int AmmoMax { get; set; }
```
mzegar, line 22: `MaxAmmo`? I'm fine with both though.

code/items/weapons/TTTWeapon.cs
```diff
@@ -157,6 +157,24 @@ public override void Simulate(Client owner)
                 }
             }
 
+            //No idea where else to put this. So it goes here.
```
mzegar, line 160: We can remove this comment, this seems like a reasonable place to put this.

code/items/weapons/TTTWeapon.cs
```diff
@@ -157,6 +157,24 @@ public override void Simulate(Client owner)
...
+            {
+                if (IsServer)
+                {
+                    var ammoBox = new BuckshotAmmo()
+                    {
+                        AmmoAmount = AmmoClip,
+                        Position = Owner.EyePos + Owner.EyeRot.Forward * 50,
+                        Rotation = Owner.EyeRot
+                    };
+                    ammoBox.Velocity = Owner.EyeRot.Forward * 500;
```
mzegar, line 171: Can we change the `500` to a const static int at the top of the file? and give it an appropriate name.

code/items/weapons/TTTWeapon.cs
```diff
@@ -157,6 +157,24 @@ public override void Simulate(Client owner)
...
 
+            //No idea where else to put this. So it goes here.
+            if (Input.Released(InputButton.View) && AmmoClip > 0) //Todo: Move this to be a bindable key (default Z???)
+            {
+                if (IsServer)
+                {
+                    var ammoBox = new BuckshotAmmo()
+                    {
+                        AmmoAmount = AmmoClip,
+                        Position = Owner.EyePos + Owner.EyeRot.Forward * 50,
```
mzegar, line 168: Can we change the `50` to a const static int at the top of the file? and give it an appropriate name.

mzegar (commented):
Here are some instructions to run the linter
https://github.com/TTTReborn/tttreborn/blob/master/CONTRIBUTING.md#setup-your-workspace

Thanks again for contributing.

code/items/ammo/BuckshotAmmo.cs
Fires1: It cannot.

The goal with using `string` rather than an enum allows for expandability. Meaning additional ammo types (especially those coming from custom addons) will easily be recognized and placed into the ammo dictionary within your ammo inventory. If it returns to being an `enum`, new ammo types cannot be added without editing the enum within the main TTTReborn gamemode's source. Having to manage all that and keep it up to date is an unnecessary piece of maintenance to add for future folk.

code/items/ammo/TTTAmmo.cs
Fires1: All inputs into the ammo inventory system are `ToLower`ed. Will stop issues of new ammo types being created if someone writes `'SMG'` rather than `'smg'`. With how it is currently, capitalization coming from the Weapon and Ammo definitions are all lowered and kept consistent in that manner.

I can't imagine there would be a need for an ammo type with different capitalization that cannot just be called something else instead. Less confusing if case sensitivity is not a factor.

code/items/ammo/BuckshotAmmo.cs
mzegar: Yeah kk that's what I assumed, thanks. I haven't really written any of the `extensible` stuff.

code/items/weapons/TTTWeapon.cs
mzegar: @Alf21 probably has a better idea than I

code/items/ammo/TTTAmmo.cs
Fires1: 50-50. I think Type works better since the entity definition is for the physical "ammo box" and it's more so picking the "type" of ammo it holds, rather than the "name" of it.
Up for consideration since the implementations are technically the only data pieces we have for certain ammo types, such as `AmmoMax` being the maximum amount of ammo a player can carry.

code/items/weapons/TTTWeapon.cs
Fires1: Up for variable name suggestions. `AmmoDropPosition`? `PositionOffset`?

mzegar: `AmmoDropPositionOffset`

code/items/ammo/TTTAmmo.cs
mzegar: I think we actually use `Name` for when referencing `Perks`, check out `TTTPerk.cs`. There might be two things here to consider

1. We might want to rename to `AmmoName`
2. We might want to drop the `Ammo` prefix as well, since people will be referencing these params as `ammo.AmmoType`

Thoughts @Alf21

Alf21: Agreed with last mentioned hint: `Name` should be the way to go, as references gonna be `ammo.Name`

code/items/weapons/TTTWeapon.cs
mzegar: Is it because `AmmoAmmount` isn't a networked variable?

code/items/ammo/TTTAmmo.cs
Fires1: Reasonable to me. Will get this all pushed in.

code/rounds/PreRound.cs
```diff
@@ -24,14 +26,32 @@ protected override void OnStart()
...
+                        ammo.Delete();
+                    }
+                    if (entity is TTTAmmoRandom rammo)
+                    {
+                        randomAmmo.Add(rammo); //Throws `Collection was Modified` if we activate here. Worth looking further into cleanup wise.
+                    }
+                    if (entity is TTTWeaponRandom rwep)
+                    {
+                        randomWeapons.Add(rwep); //See above comment.
+                    }
```
Fires1, line 49: Taking suggestions for speeding this process up if possible. May have questionable performance issues in the future with lots of entity processing.

Fires1:
Just pushed another commit with random weapon and ammo entities for Hammer.

Fires1:
Should be just about everything I want to add in terms of inventory system at this point.

Could still use a sound on pickup, but that's minor. Unsure about how to handle the collisions and being able to climb with ammo boxes. Animation for pickup is a polish piece that can come later. Same with the notification.

Best if we can get some time and test with other players, make sure ammo boxes dropped network all the data across properly and everything is in working order. Few commits still need a review so let me know 👍

Alf21:
> Should be just about everything I want to add in terms of inventory system at this point.
> 
> Could still use a sound on pickup, but that's minor. Unsure about how to handle the collisions and being able to climb with ammo boxes. Animation for pickup is a polish piece that can come later. Same with the notification.
> 
> Best if we can get some time and test with other players, make sure ammo boxes dropped network all the data across properly and everything is in working order. Few commits still need a review so let me know 👍

Gonna give you my review on 2h :) Sadly, we are unable to test until sbox fixed a bug that occurs with projects that have a large code base.

code/items/ammo/TTTAmmo.cs
```diff
@@ -0,0 +1,99 @@
...
+            SetupPhysicsFromModel(PhysicsMotionType.Dynamic);
+
+            AmmoEntMax = Amount;
+            CurrentAmmo = Amount;
+        }
+
+        public override void Touch(Entity other)
+        {
+            base.Touch(other);
+            if (IsServer)
```
Alf21, line 51:
```suggestion

            if (IsServer)
```

code/items/ammo/SMGAmmo.cs
```diff
@@ -0,0 +1,13 @@
+using Sandbox;
+
+namespace TTTReborn.Items
+{
+    [Library("ttt_ammo_smg")]
+    partial class SMGAmmo : TTTAmmo
+    {
+        public override string Type => "smg";
```
Alf21, line 8: Do we really need Type? Can't we use `SMGAmmo` or "SMGAmmo.Name" (which is set to the name of the lib) to avoid redundancy and string issues?

code/items/ammo/TTTAmmo.cs
```diff
@@ -0,0 +1,99 @@
...
+
+namespace TTTReborn.Items
+{
+    [Library("ttt_ammo")]
+    public abstract partial class TTTAmmo : Prop
+    {
+        /// <summary>
+        /// String definition of ammo type, should match TTTWeapon.AmmoType
+        /// </summary>
+        public virtual string Type { get; set; }
```
Alf21, line 15: I'd name this `Name`.

code/items/ammo/TTTAmmo.cs
```diff
@@ -0,0 +1,99 @@
...
+
+        public override void Touch(Entity other)
+        {
+            base.Touch(other);
+            if (IsServer)
+            {
+                if (other is TTTPlayer player)
+                {
+                    var ammoType = Type.ToLower();
+                    var playerInv = player.Inventory as Inventory;
```
Alf21, line 56: `inventory` instead of `playerInv` to keep it consistent and obvious :D

code/items/ammo/TTTAmmoRandom.cs
```diff
@@ -0,0 +1,24 @@
...
+using Sandbox;
+
+namespace TTTReborn.Items
+{
+    [Library("ttt_ammo_random")]
+    public class TTTAmmoRandom : Entity
+    {
+        public void Activate()
+        {
+            var ammoTypes = Globals.Utils.GetTypes<TTTAmmo>();
```
Alf21, line 12: `var` :D

code/player/Inventory.cs
```diff
@@ -107,6 +108,23 @@ public bool IsCarryingType(Type t)
...
+            {
+                if (entity is TTTWeapon wep)
+                {
+                    if (!types.Contains(wep.AmmoType))
+                    {
+                        types.Add(wep.AmmoType);
+                    }
+                }
+            }
+            return types;
```
Alf21, line 125:
```suggestion

            return types;
```

code/rounds/PreRound.cs
Alf21: Hmm yea, what are you exactly doing here? You're looping through all Entities (Existing entities), adding them to a list, and then calling `Activate` for each item, that (in the Activate method) loops through all weapons again and spawns a random weapon in top of the already existing Entity? That's a lot of entities. Instead, these entities should get replaced, minimum. Also, you could store a list of available weapons globally, initializing it on Hotreload or Init.

Alf21 (changes requested):
I really really like your changes ❤️ 

As I expected and told you, lot of code-style issues, easy to fix (don't use `var`, use empty line above `if`-clauses and before `returns, continues, break` as well as do not use one-line `if`-clauses without brackets (to cut it down).

code/items/ammo/TTTAmmo.cs
```diff
@@ -0,0 +1,99 @@
...
+
+                    if (playerInv.GetAmmoTypes().Contains(ammoType)) //Possible optimization point by caching weapons that can our ammo, rather than checking if player has our ammo.
```
mzegar, line 58: When this is about to be merged, write up an issue on github for this comment instead of leaving it as a comment.

code/items/ammo/TTTAmmo.cs
```diff
@@ -0,0 +1,99 @@
...
+                        }
+                    }
+                }
+            }
+        }
+
+        public void SetCurrentAmmo(int ammo)
+        {
+            CurrentAmmo = ammo;
+        }
```
mzegar, line 82: Is there a way to avoid doing this? Passing this data through constructor?

code/items/ammo/TTTAmmo.cs
```diff
@@ -0,0 +1,99 @@
...
+            //Unsure how Prop.Invulnerable works in comparison to this. Might be a better solution.
+            var body = info.Body;
+            if (!body.IsValid())
+                body = PhysicsBody;
+
+            if (body.IsValid() && !info.Flags.HasFlag(DamageFlags.PhysicsImpact))
+            {
+                body.ApplyImpulseAt(info.Position, info.Force * 100);
+            }
+            return;
```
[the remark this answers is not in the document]
mzegar: We should test this whenever we can, I think I ended up killing myself walking into an ammo box. We should just make the ammo box have no collision, wasn't it like that on the original TTT?

code/items/ammo/TTTAmmoRandom.cs
```diff
@@ -0,0 +1,24 @@
...
+
+            if (ammoTypes.Count > 0)
+            {
+                var typeToSpawn = ammoTypes[new Random().Next(ammoTypes.Count)];
```
mzegar, line 16: var

code/items/weapons/TTTWeapon.cs
```diff
@@ -157,6 +160,22 @@ public override void Simulate(Client owner)
...
+                    var ammoBox = new BuckshotAmmo()
+                    {
+                        Position = Owner.EyePos + Owner.EyeRot.Forward * AmmoDropPositionOffset,
+                        Rotation = Owner.EyeRot
+                    };
+                    ammoBox.Velocity = Owner.EyeRot.Forward * AmmoDropVelocity;
+                    ammoBox.SetCurrentAmmo(AmmoClip);
+                }
+
+                TakeAmmo(AmmoClip);
```
mzegar, line 176: We need separate implementations for each weapon right, we don't drop `Buckshot` for every single `TTTWeapon`. Wrap this up in a nice function and implement it for each weapon.

code/items/weapons/TTTWeapon.cs
```diff
@@ -157,6 +160,22 @@ public override void Simulate(Client owner)
                 }
             }
 
+            if (Input.Released(InputButton.View) && AmmoClip > 0) //Todo: Move this to be a bindable key (default Z???)
```
mzegar, line 163: This comment can be removed when about to be merged, should make an issue for it on our github. Or add onto the current `ammosystem` issue

mzegar (commented):
Around round of reviews, thanks again for contributing.

code/items/weapons/TTTWeapon.cs
Fires1: Totally forgot about this when I started adding new ammo, whoops. This is why we review! Will get that adjusted.

code/rounds/PreRound.cs
Fires1: It iterates through all the random spawning entities and stores them in a list to be activated, which then spawns a new weapon entity.
The issue is that in my testing, activating it within the `foreach.. in Entity.All`, it'll throw that error `Collection was Modified` due to that new entity begin added.
Unfortunately I don't know a way to remedy this myself, which is why I left the comment. Might just make it an issue and leave it for another day.

code/items/ammo/TTTAmmo.cs
Fires1: Will keep this comment open until ready to merge.

code/items/ammo/TTTAmmo.cs
```diff
@@ -0,0 +1,99 @@
...
+                if (other is TTTPlayer player)
+                {
+                    var ammoType = Type.ToLower();
+                    var playerInv = player.Inventory as Inventory;
+
+                    if (playerInv.GetAmmoTypes().Contains(ammoType)) //Possible optimization point by caching weapons that can our ammo, rather than checking if player has our ammo.
+                    {
+                        var playerAmount = playerInv.Ammo.Count(ammoType);
+
+                        if (Max >= (playerAmount + Math.Ceiling(CurrentAmmo * 0.25)))
```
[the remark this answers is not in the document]
Fires1: It follows the logic of the original implementation in Gm-TTT. Which only allows you to pickup ammo if you have space for at least a quarter of the amount of ammo within the box. Stops game unnecessarily picking up 1 or 2 bullets at a time from boxes, as well as deletes the box if it drops below 25%. Mostly a house keeping thing logically, think it best goes to make sure ammo is not randomly being picked up at weird times as well as that ammo boxes won't contain strange or odd amounts of ammo. 

I personally like how it is, I think it does a nice piece of TTT gameplay and allows you to properly rely on ammo crates as actually having a sufficient amount of ammo within them.

code/items/ammo/TTTAmmo.cs
```diff
@@ -0,0 +1,99 @@
...
+                        {
+                            var amountGiven = Math.Min(CurrentAmmo, Max - playerAmount);
+                            playerInv.Ammo.Give(ammoType, amountGiven);
+                            CurrentAmmo -= amountGiven;
+                            OnPickup.Fire(other);
+
+                            if (CurrentAmmo <= 0 || Math.Ceiling(AmmoEntMax * 0.25) > CurrentAmmo)
```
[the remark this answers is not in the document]
Fires1: See above conversation on line 62.

code/items/ammo/TTTAmmo.cs
Fires1: That's the weird function I started a conversation about earlier. Previously when trying to pass data through an inline constructor, it wouldn't take the data unless I explicitly set it with a function.
[That conversation was here, will send a response there in a moment.](https://github.com/TTTReborn/tttreborn/pull/167#discussion_r680429901)

code/items/ammo/TTTAmmo.cs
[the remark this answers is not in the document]
Fires1: It was. I messed around with a few collision override stuff and wasn't able to get a proper result. Unsure how in S&box to remove a collider from pushing a player while still tripping the `Touch` function. Open for interpretation.

code/items/weapons/TTTWeapon.cs
Fires1: Keeping this comment open until good to merge.

code/items/weapons/TTTWeapon.cs
Fires1: Unsure. AmmoAmount cannot be set with a `Net` attribute since it's a virtual property.

Fires1: There might be a way to get around that with a different inheritance setup though.

mzegar: Let me see if I can come up with something

Fires1:
Unsure if dotnet-format through the workflow tool can automatically fix the issues. So just committed the single line formatting fix. Good for another check

code/items/ammo/TTTAmmo.cs
```diff
@@ -0,0 +1,100 @@
...
+            CurrentAmmo = ammo;
+        }
+
+        public override void TakeDamage(DamageInfo info)
+        {
+            //Excerpt from Sandbox.BasePhysics | Removes the ability to remove entity health and destroy, but will still calculate appropriate physics for damage.
+            //Unsure how Prop.Invulnerable works in comparison to this. Might be a better solution.
+            var body = info.Body;
+            if (!body.IsValid())
+                body = PhysicsBody;
```
mzegar, line 91: We avoid `var`, and always have curly braces in our codebase.

code/items/ammo/TTTAmmo.cs
```diff
@@ -0,0 +1,100 @@
...
+
+        public void SetCurrentAmmo(int ammo)
+        {
+            CurrentAmmo = ammo;
+        }
+
+        public override void TakeDamage(DamageInfo info)
+        {
+            //Excerpt from Sandbox.BasePhysics | Removes the ability to remove entity health and destroy, but will still calculate appropriate physics for damage.
+            //Unsure how Prop.Invulnerable works in comparison to this. Might be a better solution.
```
mzegar, line 88: Is this comment still needed? https://github.com/TTTReborn/tttreborn/pull/167/files#r682210354

code/player/Inventory.cs
```diff
@@ -107,6 +108,26 @@ public bool IsCarryingType(Type t)
             return List.Any(x => x.GetType() == t);
         }
 
+        public IList<string> GetAmmoTypes()
+        {
+            List<string> types = new();
+
+            foreach (Entity entity in List)
+            {
+
```
mzegar, line 117: Are these spaces? seem a bit odd.

code/player/Inventory.cs
```diff
@@ -107,6 +108,26 @@ public bool IsCarryingType(Type t)
...
+                if (entity is TTTWeapon wep)
+                {
+
```
mzegar, line 120: Another odd space

mzegar (commented):
I got nothing else on my end, if @Alf21 can give it another review and we can push this PR.

code/player/Inventory.cs
```diff
@@ -107,6 +108,24 @@ public bool IsCarryingType(Type t)
             return List.Any(x => x.GetType() == t);
         }
 
+        public IList<string> GetAmmoTypes()
```
Alf21, line 111: Should be placed into `Inventory.Ammo.cs` maybe as `GetTypes()`

Alf21 (approved):
As already said on Discord, I really like your contribution. Maybe we gonna work through the code trying to find a more stable way to handle ammo without redundancies like using multiple name strings etc. 

Keep up your great work ❤️

PR #170, opened 2021-08-01
UI Menu: Added `PanelContent` and dummy data

Alf21:
image


mzegar:
@Alf21 Can you fix the merge conflicts? then I'll give it a review? Since it is showing files in the PR that are already in the codebase.

Alf21:
> @Alf21 Can you fix the merge conflicts? then I'll give it a review? Since it is showing files in the PR that are already in the codebase.

Already on it

code/ui/generalhud/menu/Menu.cs
```diff
@@ -1,21 +1,68 @@
...
     public partial class Menu : RichPanel
     {
         public static Menu Instance;
 
-        private DraggableHeader _draggableHeader;
+        private MenuHeader _menuHeader;
+
+        public readonly PanelContent MenuContent;
+
+        private readonly string _baseTitle = "Main Menu";
```
mzegar, line 14: I would personally just make this an empty string by default.

code/ui/panelcontent/PanelContent.cs
```diff
@@ -0,0 +1,129 @@
...
+            }
+
+            _historyIndex--;
+
+            UpdatePanelContent();
+        }
+
+        public void Next()
+        {
+            if (_historyIndex > _contentHistory.Count - 2)
```
mzegar, line 92: `-2` seems a bit odd here, is it because we need to subtract `-1 from both?

code/ui/generalhud/menu/Menu.cs
Alf21: So that we don't have a title?

code/ui/panelcontent/PanelContent.cs
Alf21: Because we need to check that we are not at the last index (one before the last index, and as Counts return the index+1, we have to do `-2`)

code/ui/generalhud/menu/Menu.cs
mzegar: Yeah, if no header is provided, no title is given. That seems a bit more intuitive than just to make it "Main menu" imo.

mzegar (approved):
Don't have much else, nice stuff.

PR #174, opened 2021-08-02
UI Components: Added switch component

Alf21:
Switcher:

https://user-images.githubusercontent.com/11843790/127926449-599b1c0d-56cd-4a9a-b12a-75457e82ccc1.mp4

Drag and Drop:

https://user-images.githubusercontent.com/11843790/128098926-bf80f56b-0c19-417f-9a3b-cdb5529ff4aa.mp4


code/global/Utils.cs
```diff
@@ -157,5 +158,43 @@ public static float SourceUnitsToMeters(float sourceUnits)
         {
             return sourceUnits / 39.37f;
         }
+
+        public static T GetHoveringPanel<T>(Panel excludePanel, Panel rootPanel = null) where T : Panel
```
mzegar, line 162: Would it be worth making this param a List instead? So we can provide a list of panels to exclude?

code/ui/components/dragdrop/Drag.cs
```diff
@@ -0,0 +1,239 @@
...
+                        _oldPositionMode = DragBasePanel.Style.Position;
+                    }
+
+                    DragBasePanel.Style.Position = PositionMode.Absolute;
+                }
+            }
+        }
+        private bool _isDragging = false;
+
+        public bool IsFreeDraggable
```
mzegar, line 38: What does this `bool` mean? That the component can be dragged around? Not sure if I understand.

Alf21: That the component can be dragged without limitation (it does not need to snap into the parent, it can stay where you dragged it)

mzegar: sounds fine then, thanks.

code/global/Utils.cs
Alf21: If needed, we can do. But currently, even not this function is in use. By default, Panel's Mouse events should be used (but these wouldn't work for panels, that's why I created it - then I came up with a better way, that is not as extensible tho).

mzegar (approved):
LGTM

PR #175, opened 2021-08-02
Fea: Add detective role

mzegar:


Alf21:
I think instead of assigning roles in a somehow bad way to the items, it would be better creating a shopping system and assigning items to shops and shops to roles instead :)

mzegar:
Closing this for now, going to rethink this approach and implementation.

PR #177, opened 2021-08-04
Added `Dropdown`, `Modal`, `DialogBox`, `Keybind` and `FileSelection` UI component

Alf21:
DropDown:

https://user-images.githubusercontent.com/11843790/128244537-a3bf0563-2c15-4fec-a075-e45a44c4504c.mp4

image

image

### DIFF
--- a/code/ui/components/dropdown/Dropdown.cs
+++ b/code/ui/components/dropdown/Dropdown.cs
@@ -1,0 +1,62 @@
+using System;
+using System.Collections.Generic;
+
+using Sandbox.UI;
+using Sandbox.UI.Construct;
+
+namespace TTTReborn.UI
+{
+    public partial class Dropdown : TTTPanel
+    {
+        public readonly Label TextLabel;
+
+        public bool IsOpened
+        {
+            get => OptionHolder.IsShowing;
+            private set
+            {
+                OptionHolder.IsShowing = value;
+
+                _openLabel.SetClass("opened", value);
+            }
+        }
+
+        private readonly Label _openLabel;
+
+        public readonly List<DropdownOption> Options = new();
+
+        public readonly TTTPanel OptionHolder;
+
+        public Dropdown(Panel parent = null) : base(parent)
+        {
+            Parent = parent ?? Parent;
+
+            StyleSheet.Load("/ui/components/dropdown/Dropdown.scss");
+
+            TextLabel = Add.Label("", "textLabel");
+            _openLabel = Add.Label("expand_more", "openLabel");
+
+            OptionHolder = new TTTPanel(this);
+            OptionHolder.AddClass("optionholder");
+        }
+
+        public void AddOption(string text, Action<TTTPanel> onSelect)
+        {
+            Options.Add(new DropdownOption(OptionHolder, text)
+            {
+                Dropdown = this,
+                OnSelect = onSelect
+            });
+        }
+
+        public virtual void OnSelectOption(DropdownOption option)
+        {
+            IsOpened = false;
+        }
+
+        protected override void OnClick(MousePanelEvent e)
+        {
+            IsOpened = !IsOpened;
+        }
+    }
+}
--- a/code/ui/components/dropdown/Dropdown.cs
+++ b/code/ui/components/dropdown/Dropdown.cs
@@ -27,31 +27,47 @@ namespace TTTReborn.UI
 
         public readonly TTTPanel OptionHolder;
 
+        public DropdownOption SelectedOption
+        {
+            get => _selectedOption;
+            private set
+            {
+                _selectedOption = value;
+
+                TextLabel.Text = _selectedOption.TextLabel.Text;
+            }
+        }
+        private DropdownOption _selectedOption;
+
         public Dropdown(Panel parent = null) : base(parent)
         {
             Parent = parent ?? Parent;
 
             StyleSheet.Load("/ui/components/dropdown/Dropdown.scss");
 
-            TextLabel = Add.Label("", "textLabel");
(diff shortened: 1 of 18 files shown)


code/ui/components/dropdown/Dropdown.cs
```diff
@@ -0,0 +1,62 @@
...
+using Sandbox.UI;
+using Sandbox.UI.Construct;
+
+namespace TTTReborn.UI
+{
+    public partial class Dropdown : TTTPanel
+    {
+        public readonly Label TextLabel;
+
+        public bool IsOpened
```
mzegar, line 13: `IsOpen`, `Opened` is past tense. Unless you wanted to say `WasOpened`

Alf21:
Wasn't able to find a `Dropdown` component in `Sandbox.UI`

code/ui/components/dropdown/DropdownOption.cs
```diff
@@ -0,0 +1,29 @@
...
+            TextLabel = Add.Label(text ?? "Select", "optiontext");
+        }
+
+        protected override void OnClick(MousePanelEvent e)
+        {
+            OnSelect?.Invoke(this);
+            Dropdown?.OnSelectOption(this);
+        }
+    }
+}
```
mzegar, line 29: Imo this can be in the same file as `Dropdown.cs`

code/ui/components/dropdown/DropdownOption.cs
```diff
@@ -0,0 +1,29 @@
...
+{
+    public partial class DropdownOption : TTTPanel
+    {
+        public Dropdown Dropdown { get; set; }
+
+        public Label TextLabel { get; private set; }
+
+        public Action<TTTPanel> OnSelect { get; set; }
+
+        public DropdownOption(Panel parent = null, string text = null) : base(parent)
```
mzegar, line 16: Can we just do empty string as a default instead of null for `text`?

code/ui/components/keybind/Keybind.cs
```diff
@@ -0,0 +1,61 @@
...
+                dialogBox.OnDecline = (panel) => panel.Close();
+                dialogBox.Display();
+            }
+        }
+    }
+}
+
+namespace Sandbox.UI.Construct
+{
+    using TTTReborn.UI;
```
mzegar, line 49: Double namespaces?

code/ui/components/dropdown/DropdownOption.cs
Alf21: Wouldn't put it together, as we have a folder anyways.

code/ui/components/dropdown/DropdownOption.cs
Alf21: And what's the benefit? <https://github.com/TTTReborn/tttreborn/pull/177/files#diff-c74c736f039bae78fc73d7b0f9b420ba7a81707b4c27469924b1d550e428dde2R20>

code/ui/components/keybind/Keybind.cs
Alf21: This is needed as we're in a different namespace

code/ui/components/dropdown/DropdownOption.cs
AMastryukov: It's generally better to use String.Empty as it's just safer overall since you won't get null refs.
It's just better practice as far as I'm aware.

code/ui/tttpanel/TTTPanel.cs
```diff
@@ -27,27 +27,26 @@ public void AddChild(Panel child, int index)
...
-                while (index < childCount)
-                {
-                    _children[index] = tmp;
-                    tmp = _children[++index];
-                }
-
-                _children[index] = tmp;
-                _children[++index] = null;
-            }
+            // int childCount = ChildCount;
```
AMastryukov, line 30: Do we need to keep this commented out?
Could we just remove it?

AMastryukov (approved):
Looks good overall, I'd remove the commented stuff out unless it's absolutely needed to remain in there for the future, but then it should have a comment explaining why it's there.

code/ui/components/dropdown/DropdownOption.cs
mzegar: ^ yeah just better practice basically, if you have the option to use something empty, you should.

code/ui/components/keybind/Keybind.cs
mzegar: just making sure : D

code/ui/components/dropdown/DropdownOption.cs
mzegar: no worries, just a suggestion.

mzegar:
IMO go with the empty string, otherwise thanks for all the work.

code/ui/components/dropdown/Dropdown.cs
```diff
@@ -0,0 +1,62 @@
...
+        }
+
+        public void AddOption(string text, Action<TTTPanel> onSelect)
+        {
+            Options.Add(new DropdownOption(OptionHolder, text)
+            {
+                Dropdown = this,
+                OnSelect = onSelect
+            });
+        }
```
mzegar, line 50: You'll probably also want to provide a default value for `text` here, since `DropdownOption()` is the one with the default param.

I would personally go with empty string for both.

code/ui/tttpanel/TTTPanel.cs
Alf21: I need this for upcoming child repositioning, currently, it keeps crashing the game, that's why it's commented out

PR #178, opened 2021-08-05
Core: Added translation system

pBinarySoul:
Simple translation system.

image


code/lang/packs/English.json
```diff
@@ -0,0 +1,40 @@
...
+
+    "POST_ROUND_WIN_INNOCENT": "THE INNOCENT WIN!",
+    "POST_ROUND_WIN_TRAITORS": "THE TRAITORS WIN!",
+    "POST_ROUND_TEXT": "Thanks for playing TTT Reborn, more updates and stats to come!",
+
+    "PLAYER_STATUS_HEALTH_HEALTHY": "Healthy",
+    "PLAYER_STATUS_HEALTH_HURT": "Hurt",
+    "PLAYER_STATUS_HEALTH_WOUNDED": "Wounded",
+    "PLAYER_STATUS_HEALTH_BADLY_WOUNDED": "Badly Wounded",
+    "PLAYER_STATUS_HEALTH_NEAR_DEATH": "Near Вeath",
```
Alf21, line 20: `Beath`

code/lang/packs/English.json
```diff
@@ -0,0 +1,40 @@
+{
```
Alf21, line 1: Can't we get rid of these outer brackets, they are really annoying and could lead to code inconsistencies and issues in contributions (tab/space indentation).

Then it would be the perfect system, as it was planned to be 😃

code/ui/generalhud/gametimer/GameTimer.cs
```diff
@@ -37,10 +41,29 @@ public override void Tick()
                     return;
                 }
 
+                lang = ILanguage.GetActiveLanguage();
```
Alf21, line 44: For now it's okay, but I think I gonna rework this a bit with own text labels and language change refreshes so we can handle them with a good performance :)

code/ui/generalhud/gametimer/GameTimer.cs
```diff
@@ -37,10 +41,29 @@ public override void Tick()
...
+                        break;
+
+                    case Rounds.InProgressRound:
+                        _textLabel.Text = lang.GetTranslation("ROUND_STATE_IN_PROGRESS");
+                        break;
+
+                    case Rounds.PostRound:
+                        _textLabel.Text = lang.GetTranslation("ROUND_STATE_POST_ROUND");
+                        break;
+                }
```
Alf21, line 65: That could be handled with RoundName.Upper() concat, but I'm okay with that solution as well as there won't be more round types

code/ui/generalhud/gametimer/GameTimer.scss
```diff
@@ -1,32 +1,32 @@
...
+    border-radius: 0 0 4px 4px;
 
-    left: 50%;
-    transform: translateX(-50%);
     position: absolute;
     top: 0;
+    left: 50%;
+    transform: translateX(-50%);
+
+    width: 250px;
```
Alf21, line 12: `min-width` maybe would be better for long words

code/ui/generalhud/playerinfo/PlayerInfo.cs
```diff
@@ -65,8 +69,15 @@ public override void Tick()
...
-
+                    switch (_currentRole) {
+                        case InnocentRole:
+                            _roleLabel.Text = lang.GetTranslation("ROLE_NAME_INNOCENT");
+                            break;
+
+                        case TraitorRole:
+                            _roleLabel.Text = lang.GetTranslation("ROLE_NAME_TRAITOR");
+                            break;
+                    }
```
Alf21, line 80: That definitely should be handled with `$"ROLE_NAME_{_currentRole.Name.ToUpper()}"` or similar OR in the role instead of here :)

code/ui/generalhud/postroundmenu/PostRoundMenu.cs
```diff
@@ -58,8 +63,21 @@ public void OpenAndSetPostRoundMenu(PostRoundStats stats)
...
+            switch (_stats.WinningRole)
+            {
+                case "Innocents":
+                    _headerLabel.Text = lang.GetTranslation("POST_ROUND_WIN_INNOCENT");
+                    break;
+
+                case "Traitors":
+                    _headerLabel.Text = lang.GetTranslation("POST_ROUND_WIN_TRAITORS");
+                    break;
+            }
```
Alf21, line 79: Here should the team be used in the same way as suggested above

code/lang/ILanguage.cs
```diff
@@ -0,0 +1,58 @@
...
+        {
+            TLanguage lang = null;
+            if (!Languages.TryGetValue(name, out lang))
+            {
+                Log.Info($"Tried to get a language that doesn't exist: {name}.");
+
+                if (Host.IsServer ? ServerLanguage == name : ActiveLanguage == name)
+                    FallbackLanguage();
+            }
+            return lang;
```
Alf21, line 55:
```suggestion

            return lang;
```

code/lang/ILanguage.cs
```diff
@@ -0,0 +1,58 @@
...
+            FallbackLanguage();
+        }
+
+        public static TLanguage GetActiveLanguage() => GetLanguageByName( Host.IsServer ? ServerLanguage : ActiveLanguage );
+        public static TLanguage GetLanguageByName(string name)
+        {
+            TLanguage lang = null;
+            if (!Languages.TryGetValue(name, out lang))
+            {
+                Log.Info($"Tried to get a language that doesn't exist: {name}.");
```
Alf21, line 50: Warning would be better

code/lang/ILanguage.cs
```diff
@@ -0,0 +1,58 @@
...
+
+                if (Host.IsServer ? ServerLanguage == name : ActiveLanguage == name)
+                    FallbackLanguage();
```
Alf21, line 53: Missing surrounding brackets

code/lang/ILanguage.cs
```diff
@@ -0,0 +1,58 @@
+using System;
+using System.IO;
+using System.Collections.Generic;
+using Sandbox;
+
+namespace TTTReborn.Language
+{
+    public interface ILanguage
```
Alf21, line 8: Pretty uncommon to use Interfaces like this (as classes), I don't see you adding them to any class. I think, we would go better with a (partial?) class here

code/lang/ILanguage.cs
```diff
@@ -0,0 +1,58 @@
+using System;
+using System.IO;
+using System.Collections.Generic;
+using Sandbox;
```
Alf21, line 4:
```suggestion

using Sandbox;
```

code/lang/ILanguage.cs
```diff
@@ -0,0 +1,58 @@
...
+
+        private static void FallbackLanguage()
+        {
+            //Log.Info($"Falling back to {FALLBACK_LANGUAGE}.");
+
+            if (Host.IsClient)
+                ActiveLanguage = FALLBACK_LANGUAGE;
+
+            if (Host.IsServer)
+                ConsoleSystem.SetValue("ttt_server_language", FALLBACK_LANGUAGE);
```
Alf21, line 28: Missing surrounding brackets, can be combined with an `else` as well

code/lang/ILanguage.cs
```diff
@@ -0,0 +1,58 @@
...
+        }
+        public static void LoadLanguages()
```
Alf21, line 30:
```suggestion

        public static void LoadLanguages()
```

code/lang/ILanguage.cs
```diff
@@ -0,0 +1,58 @@
...
+                string json = FileSystem.Mounted.ReadAllText($"/lang/packs/{file}");
+                Languages.Add(name, new TLanguage(name, json));
+
+                Log.Info($"Added TTT language pack: {name}.");
+            }
+
+            FallbackLanguage();
+        }
+
+        public static TLanguage GetActiveLanguage() => GetLanguageByName( Host.IsServer ? ServerLanguage : ActiveLanguage );
```
Alf21, line 44:
```suggestion
        public static TLanguage GetActiveLanguage() => GetLanguageByName(Host.IsServer ? ServerLanguage : ActiveLanguage);

```

code/lang/ILanguage.cs
```diff
@@ -0,0 +1,58 @@
...
+        public static TLanguage GetLanguageByName(string name)
+        {
+            TLanguage lang = null;
+            if (!Languages.TryGetValue(name, out lang))
```
Alf21, line 48:
```suggestion
            if (!Languages.TryGetValue(name, out TLanguage lang))
```

code/lang/ILanguage.cs
```diff
@@ -0,0 +1,58 @@
...
+{
+    public interface ILanguage
+    {
+        // TODO: On cvar change callback
+        [ClientVar("ttt_language", Help = "Your current language", Saved = true)]
+        public static string ActiveLanguage { get; set; }
+
+        [ServerVar("ttt_server_language", Help = "Server's current language", Saved = true)]
+        public static string ServerLanguage { get; set; }
+        public const string FALLBACK_LANGUAGE = "English";
```
Alf21, line 16: I'd like to avoid these language names and instead using "enUS" or other ISO styles to support region based language specifications as well :)

Maybe we should add a json header / config for that language as well then (original language name)? Then we should keep the outer brackets of the json for sure.

Alf21 (changes requested):
I really like what you did, a core system and really good and simple-kept solution, so translation contributors wouldn't have to struggle.

code/lang/packs/English.json
Fires1: Not sure if I'm understanding you totally correctly Alf, but I'm pretty sure those are required within JSON as curly brackets are the 'object' markers in order for the parser to read it. Could be mistaken.

Alf21: Yea yea, for sure (even that there are different possibilities out there), but to simplify the creation, we could easily add brackets at start and end before reading the data ^^

Anyways, if we wanna add configs to the json as well, then these brackets are needed tho

Fires1: Reasonable. Could alternatively swap out Json for just reading line by line and parsing the individual lines through our own code. Might be a tad bit faster than the Json reader.

I'd support having the file reader check to see if the brackets exist at the start and end within the file, then add them if they do not, so it works whether you have them or not.

code/ui/generalhud/playerinfo/PlayerInfo.cs
mzegar: It might be worth just have methods inside of the ILanguage class that handles this logic.

I.e

```csharp
string GetRoleTranslation(Role role) {
	...
}
```

Thoughts?

code/lang/ILanguage.cs
mzegar: Agreed, an interface shouldn't be used here. Could just go with a static class.

mzegar (changes requested):

code/ui/generalhud/gametimer/GameTimer.cs
mzegar: What are your ideas with handling this? I think we should just have a single `language` object that exists either on the top-level hud or somewhere on the game instance, instead of getting the active language for each UI element.

In terms of handling the language change, it could be as simple as refreshing the hud.

code/lang/ILanguage.cs
pBinarySoul: I thought about that. But in this case I had to rewrite TLanguage constructor. :D

code/ui/generalhud/playerinfo/PlayerInfo.cs
pBinarySoul: @mzegar
With English - yes. But, for example, in Russian you can't just say "Он был Предатель", instead you say "Он был Предател**ем**"

pBinarySoul: $"ROLE_NAME_{_currentRole.Name.ToUpper()}" is more appropriate. But I think switch is faster and more readable.

code/ui/generalhud/gametimer/GameTimer.cs
pBinarySoul: @Alf21
First I need to learn how to track cvar changes.

code/lang/ILanguage.cs
AMastryukov: Languages could be specified by an enum with a single invariant culture language (English) that specifies keys for localized string lookups.

code/lang/packs/English.json
pBinarySoul: ¯\_(ツ)_/¯

code/ui/generalhud/postroundmenu/PostRoundMenu.cs
mzegar: Is it possible for other roles to win? We might want to do something similar to 
`public string GetRoleTranslation(string key)`
so that we can support different roles winning the game.

mzegar (commented):
Only one more comment on my end, otherwise thanks again for the contribution.

code/ui/generalhud/postroundmenu/PostRoundMenu.cs
Alf21: Even if not planned on our side, it should be possible :)

code/lang/ILanguage.cs
Alf21: > Languages could be specified by an enum with a single invariant culture language (English) that specifies keys for localized string lookups.

That's not a good idea as language files should be loaded on fly instead of needing to interact with the API / C# code directly.

Maybe we should create a default folder structure, so all files inside a lang folder that are json files are getting initialized. And that eg. a enUS.json or config.json contains the relevant data. Or just combined inside one file via json syntax (just reserve they keyword "config" e.g.)

Alf21:
I made the changes I requested on my own. ConVar saving is not enabled currently. That could be handled with setting the convar on `ttt_language` change and loading the data on initialization.

Alf21 (approved):

code/lang/TTTLanguage.cs
```diff
@@ -1,70 +1,66 @@
...
         }
+        private static Language _serverLanguage;
+
+        public static readonly Dictionary<string, Language> Languages = new();
 
         public static void LoadLanguages()
-        {            
-            foreach ( string file in FileSystem.Mounted.FindFile("/lang/packs/", "*.json", false) )
+        {
+            foreach (string file in FileSystem.Mounted.FindFile("/lang/packs/", "*.json", false))
```
pBinarySoul, line 36: I forgot to point out that this doesn't support hot loading. We need to figure out how to do this, that'll be really great.

Alf21: Afaik, hotloading won't trigger for .json files, that's why I don't think we need hotloading.

code/lang/Language.cs
```diff
@@ -0,0 +1,99 @@
...
+            if (translation is not null)
+            {
+                return translation.ToString();
+            }
+            else if (TTTLanguage.Languages.TryGetValue(TTTLanguage.FALLBACK_LANGUAGE, out Language fallbackLanguage) && fallbackLanguage != this)
+            {
+                return fallbackLanguage.GetTranslation(key);
+            }
+
+            return $"[ERROR: Translation of '{key}' not found]";
```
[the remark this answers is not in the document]
Alf21: Oh yea, you are right, feel free adding such a mechanic :)

mzegar (approved):

PR #180, opened 2021-08-06
Add TTTPanel to all Panels

Travh98:
I also set the IsShowing member variables everywhere that setClass("hide", true/false) was.
This is my first pull request and also first time using C#. I'm excited to help!

### DIFF
--- a/code/items/perks/radar/RadarPoint.cs
+++ b/code/items/perks/radar/RadarPoint.cs
@@ -7,7 +7,7 @@ using TTTReborn.UI;
 
 namespace TTTReborn.UI
 {
-    public class RadarPoint : Panel
+    public class RadarPoint : TTTPanel
     {
         public Vector3 Position;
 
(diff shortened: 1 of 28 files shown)


mzegar:
Hey, thanks again for the PR.

Check out `InspectMenu.cs` there is an implementation for `IsShowing` we can remove, since we use the TTT Panel.

mzegar:
It might be good to investigate why the inventory was not showing after it inherits from `TTTPanel`, because I think we would like to make `TTTPanel` a standard for all UI elements.

Travh98:
Thank you for the review, I made the changes you mentioned.
As for the issue with Inventory not showing, I'll open a separate branch and troubleshoot that there.

mzegar (approved):
LGTM

mzegar:
Spoke with Travh, figured out the issue with the inventory.

PR #181, opened 2021-08-06
Add various map entities and round change event

Fires1:
Finally going to get this in a draft to begin review. Recommend reading the list below before poking code to have an idea where things are at right now.

I also wrote a wiki page entry for using some of these entities, [which can be read here.](https://gist.github.com/Fires1/c559fe715318ab888f7a48f40a7f2f4b)

- Added `tttreborn.round.changed` event
- - Passes the `BaseRound round` parameter of `Game.ForceRoundChange` when fired.
- Added `ttt_map_settings` entity
- - Ironically has no settings at the moment but provides Outputs for round changes.
- Added `ttt_credit_adjust` entity
- - Identical logically to credit adjust entity from Gmod TTT. Removes/adds credits. 
- - Does not check if role is traitor or detective, can innocents technically hold credits? I do not know.
- Added `ttt_feed_entry`
- - Adds text to upper right feed. Can be sent to different people
- Added `ttt_force_win`
- - Forces a win for a certain team, can be the activator's team or a custom string input. Has built in Inputs for innocents and traitors.
- Added `ttt_logic_assigned`
- - Provides an output for the assigned team or role of the activator. Traitor tester and similiar.
- **!!! Work In Progress !!!** Added `ttt_role_button`
- - Creates a button that a certain role can use. Awaiting UI work from Westsaid. 
- - Logic is there, but would like a review of the networking ([and the 12 separate function calls in three lines](https://github.com/Fires1/tttreborn/blob/d533ee0cc857fbbd55b22b9cae693bf435ba0391/code/player/TTTPlayer.RoleButtons.cs#L16-L20))

### DIFF
--- a/code/global/RPCs.cs
+++ b/code/global/RPCs.cs
@@ -1,7 +1,7 @@
 using Sandbox;
 
 using TTTReborn.Items;
-using TTTReborn.MapEntities;
+using TTTReborn.Map;
 using TTTReborn.Player;
 using TTTReborn.Roles;
 using TTTReborn.Teams;
--- a/code/global/RPCs.cs
+++ b/code/global/RPCs.cs
@@ -1,6 +1,7 @@
 using Sandbox;
 
 using TTTReborn.Items;
+using TTTReborn.MapEntities;
 using TTTReborn.Player;
 using TTTReborn.Roles;
 using TTTReborn.Teams;
@@ -154,6 +155,12 @@ namespace TTTReborn.Globals
         public static void ClientClearInventory()
         {
             Event.Run("tttreborn.player.inventory.clear");
+        }
+
+        [ClientRpc]
+        public static void ClientDisplayMessage(string message, Color color)
+        {
+            InfoFeed.Current?.AddEntry(message);
         }
     }
 }
--- a/code/global/RPCs.cs
+++ b/code/global/RPCs.cs
@@ -160,7 +160,7 @@ namespace TTTReborn.Globals
         [ClientRpc]
         public static void ClientDisplayMessage(string message, Color color)
         {
-            InfoFeed.Current?.AddEntry(message);
+            InfoFeed.Current?.AddEntry(message, color);
         }
     }
 }
--- a/code/global/Utils.cs
+++ b/code/global/Utils.cs
@@ -6,6 +6,7 @@ using Sandbox;
 using Sandbox.UI;
 
 using TTTReborn.Player;
+using TTTReborn.Roles;
 
 namespace TTTReborn.Globals
 {
@@ -86,13 +87,13 @@ namespace TTTReborn.Globals
             return clients;
         }
 
-        public static List<TTTPlayer> GetAlivePlayersByRoleName(string role)
+        public static List<TTTPlayer> GetAlivePlayersByRoleName(TTTRole role)
         {
             List<TTTPlayer> players = new List<TTTPlayer>();
 
             foreach (Client client in Client.All)
             {
-                if (client.Pawn is TTTPlayer player && player.LifeState == LifeState.Alive && player.Role.Name.ToLower() == role.ToLower())
+                if (client.Pawn is TTTPlayer player && player.LifeState == LifeState.Alive && player.Role.Name == role.Name)
                 {
                     players.Add(player);
                 }
--- a/code/global/Utils.cs
+++ b/code/global/Utils.cs
@@ -86,6 +86,21 @@ namespace TTTReborn.Globals
             return clients;
         }
 
+        public static List<TTTPlayer> GetAlivePlayersByRoleName(string role)
+        {
+            List<TTTPlayer> players = new List<TTTPlayer>();
+
+            foreach (Client client in Client.All)
+            {
+                if (client.Pawn is TTTPlayer player && player.LifeState == LifeState.Alive && player.Role.Name.ToLower() == role.ToLower())
+                {
+                    players.Add(player);
+                }
+            }
+
+            return players;
+        }
+
         public static bool HasMinimumPlayers()
         {
             return Client.All.Count >= Gamemode.Game.TTTMinPlayers;
--- a/code/mapentities/TTTCreditAdjust.cs
+++ b/code/mapentities/TTTCreditAdjust.cs
@@ -1,0 +1,43 @@
+using Sandbox;
+
+using TTTReborn.Player;
+
+namespace TTTReborn.MapEntities
+{
+    [Library("ttt_credit_adjust", Description = "Changes the amount of credits upon the activator.")]
+    public partial class TTTChangeCredits : Entity
+    {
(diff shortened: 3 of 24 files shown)


code/mapentities/TTTLogicAssigned.cs
```diff
@@ -0,0 +1,68 @@
...
+using TTTReborn.Player;
+using TTTReborn.Rounds;
+
+namespace TTTReborn.MapEntities
+{
+    [Library("ttt_logic_assigned", Description = "Used to test the assigned team or role of the activator.")]
+    public partial class TTTLogicAssigned : Entity
+    {
+        [Property("Check Type", "Entity will only return a result for one of the select values.")]
+        public Check CheckType { get; set; } = Check.Role;
```
mzegar, line 12: `CheckType` seems a bit of an odd name here?

code/global/Utils.cs
```diff
@@ -86,6 +86,21 @@ public static IEnumerable<Client> GetClientsSpectatingPlayer(TTTPlayer player)
             return clients;
         }
 
+        public static List<TTTPlayer> GetAlivePlayersByRoleName(string role)
```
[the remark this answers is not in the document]
Fires1: Not a bad idea either. I'd still probably want to compare the lowered names though.

code/mapentities/TTTLogicAssigned.cs
Fires1: Open for suggestions.

mzegar: Yeah don't have any good suggestions, keep it as is.

code/global/Utils.cs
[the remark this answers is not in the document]
Alf21: We should - if not already done - set the role's name to lowercase by default (to avoid interaction issues with outside resources or client input). Every role name has to be unique, so it doesn't matter. Should help avoiding issues in language interaction as well.

code/global/RPCs.cs
```diff
@@ -1,6 +1,7 @@
 using Sandbox;
 
 using TTTReborn.Items;
+using TTTReborn.MapEntities;
```
Alf21, line 4: I'd use `Map` as namespace name, maybe `Entities` as sub namespace

code/global/RPCs.cs
```diff
@@ -155,5 +156,11 @@ public static void ClientClearInventory()
         {
             Event.Run("tttreborn.player.inventory.clear");
         }
+
+        [ClientRpc]
+        public static void ClientDisplayMessage(string message, Color color)
```
Alf21, line 161: Color isn't used

code/global/Utils.cs
```diff
@@ -86,6 +86,21 @@ public static IEnumerable<Client> GetClientsSpectatingPlayer(TTTPlayer player)
...
+        {
+            List<TTTPlayer> players = new List<TTTPlayer>();
+
+            foreach (Client client in Client.All)
+            {
+                if (client.Pawn is TTTPlayer player && player.LifeState == LifeState.Alive && player.Role.Name.ToLower() == role.ToLower())
```
Alf21, line 95: The `toLower` function should be run and stored outside the foreach loop

code/mapentities/TTTCreditAdjust.cs
```diff
@@ -0,0 +1,43 @@
...
+
+using TTTReborn.Player;
+
+namespace TTTReborn.MapEntities
+{
+    [Library("ttt_credit_adjust", Description = "Changes the amount of credits upon the activator.")]
+    public partial class TTTChangeCredits : Entity
+    {
+        [Property("Credits", "Removes amount of credits. Use a negative number to add credits to the activator.")]
+        public int Credits { get; set; } = 0;
```
Alf21, line 11: I wouldn't design it the way you just can remove credits, instead, a general approach would be cool (possibility to add credits as well).

code/mapentities/TTTFeedEntry.cs
```diff
@@ -0,0 +1,53 @@
...
+                    RPCs.ClientDisplayMessage(To.Everyone, Message, Color);
+                    break;
+                case "innocents":
+                    RPCs.ClientDisplayMessage(To.Multiple(TeamFunctions.GetTeam("Innocents").GetClients()), Message, Color);
+                    break;
+                case "traitors":
+                    RPCs.ClientDisplayMessage(To.Multiple(TeamFunctions.GetTeam("Traitors").GetClients()), Message, Color);
+                    break;
+                default:
+                    TTTTeam team = TeamFunctions.GetTeam(Receiver);
```
Alf21, line 39: No `toLower` on receiver here. That's what I meant, leading to inconsistencies.

code/mapentities/TTTForceWin.cs
```diff
@@ -0,0 +1,60 @@
...
+        private void ForceEndRound(TTTTeam team)
+        {
+            if (Gamemode.Game.Instance.Round is InProgressRound)
+            {
+                //Logic taken from InProgressRound.LoadPostRound. Should reference the function instead?
+                Gamemode.Game.Instance.ForceRoundChange(new PostRound());
+                RPCs.ClientOpenAndSetPostRoundMenu(
+                    team.Name,
+                    team.Color
+                );
```
Alf21, line 56: Should be combined to one function imo, serversided

code/player/TTTPlayer.RoleButtons.cs
```diff
@@ -0,0 +1,29 @@
...
+
+namespace TTTReborn.Player
+{
+    public partial class TTTPlayer
+    {
+        public static List<TTTRoleButton> RoleButtons = new();
+
+        public void SendRoleButtonsToClient() //Called Serverside when Player is assigned role.
+        {
+            var roleButtons = All.Where(x => x.GetType() == typeof(TTTRoleButton)).Select(x => x as TTTRoleButton);
```
Alf21, line 16: var

Alf21:
The wiki can be found here (for now): https://github.com/TTTReborn/tttreborn/wiki :)

code/global/Utils.cs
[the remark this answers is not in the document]
Fires1: Resolved with latest commit. Force role to auto .ToLower() and compared with the role instead

code/mapentities/TTTCreditAdjust.cs
Fires1: I prefer the logic this way as it is how it worked in the original TTT. I've gone ahead and changed the property description to be a bit more specific.

Fires1:
Waiting to hear back from Westsaid regarding some UI elements. As it stands, I might throw another attempt at it after my painful (but useful) experience with the C4.

Fires1:
I'm going to go ahead and close this, and merge my forked branch into a separate branch on this repo. Will finish off the UI work for the traitor buttons then open a PR for merging into master.

PR #186, opened 2021-08-13
Pull `Fires1/tttreborn:weapons` branch into `Fires` for development (…

AMastryukov:
…#185)

- Add body armor
- Add health kit
- Add decoy
- Adjust Preround cleanup logic
- Cleanup
- Remove collision group comment
- Slider component
- Non-implemented radio/noisemaker
- C4 model, placement logic and some more

Material business plus model taken from public domain.

Code includes a shell of a UI, the equipment code as well as the entity code for placing and networking the "OnUse" to the client for the UI management.


code/items/equipments/C4Equipment.cs
```diff
@@ -0,0 +1,99 @@
...
+            {
+                if (Input.Pressed(InputButton.Attack1)) //Drops C4
+                {
+                    _ = new C4Entity()
+                    {
+                        Position = Owner.EyePos + Owner.EyeRot.Forward * DropPositionOffset,
+                        Rotation = Owner.EyeRot,
+                        Velocity = Owner.EyeRot.Forward * DropVelocity,
+                    };
+                    (owner.Inventory as Inventory).Remove(this);
```
mzegar, line 56: Could just extract this logic into a method
`dropC4`, then you wouldn't need the comment.

code/items/equipments/C4Equipment.cs
```diff
@@ -0,0 +1,99 @@
...
+                    bomb.Rotation = bomb.Rotation.RotateAroundAxis(Vector3.Up, 90);
+
+                    if (placementTrace.Entity is TTTPlayer player)
+                    {
+                        bomb.EnableSolidCollisions = false;
+                        bomb.AttachedBone = placementTrace.Bone;
+                        bomb.SetParent(player, placementTrace.Bone);
+                    }
+
+                    (owner.Inventory as Inventory).Remove(this);
```
mzegar, line 92: Likewise with comment above, extract into a method, `plantC4`, no need for the comment. Mostly just a coding style change.

code/items/equipments/C4Equipment.cs
```diff
@@ -0,0 +1,99 @@
...
+                }
+            }
+        }
+
+        public override bool CanDrop() => false;
```
mzegar, line 97: Just a coding style thing, but I prefer these at the top of the file, but it isn't in our coding standards or anything.

code/items/equipments/HealthkitEquipment.cs
```diff
@@ -0,0 +1,62 @@
...
+            {
+                if (Input.Pressed(InputButton.Attack1))
+                {
+                    _ = new HealthkitEntity()
+                    {
+                        Position = Owner.EyePos + Owner.EyeRot.Forward * DropPositionOffset,
+                        Rotation = Owner.EyeRot,
+                        Velocity = Owner.EyeRot.Forward * DropVelocity,
+                    };
+                    (owner.Inventory as Inventory).Remove(this);
```
mzegar, line 55: I've seen this similar code three times, I wonder if we could write a utility method for it. It may end up being cleaner.

code/items/equipments/entities/C4Entity.cs
```diff
@@ -0,0 +1,150 @@
...
+        public bool IsArmed { get; set; } = false;
+
+        //Timer display on C4 entity.
+        private WorldPanel Display;
+        private bool CreatedDisplay = false;
+
+        //Explosion parameters.
+        private const int RADIUS = 1024;
+        private const int DAMAGE = 500;
+        private const int FORCE = 50;
```
mzegar, line 30: Could avoid the comment, and just write
`BOMB_RADIUS`, `BOMB_DAMAGE`, etc.

code/items/equipments/entities/C4Entity.cs
```diff
@@ -0,0 +1,150 @@
...
+
+        [Net]
+        public int AttachedBone { get; set; } = -1;
+
+        [Net]
+        public bool IsArmed { get; set; } = false;
+
+        //Timer display on C4 entity.
+        private WorldPanel Display;
+        private bool CreatedDisplay = false;
```
mzegar, line 25: Could avoid the comment, and just make the variable name clearer.

`BOMB_DISPLAY`

code/items/equipments/entities/C4Entity.cs
```diff
@@ -0,0 +1,150 @@
...
+
+namespace TTTReborn.Items
+{
+    [Library("ttt_c4_ent"), Hammer.Skip]
+    public partial class C4Entity : Prop, IUse
+    {
+        private string ModelPath => "models/entities/c4.vmdl";
+
+        [Net]
+        public int AttachedBone { get; set; } = -1;
```
mzegar, line 18: Might be a good place for a comment, why is `attachedbone` set to `-1`?

code/items/equipments/entities/C4Entity.cs
```diff
@@ -0,0 +1,150 @@
...
+            return false; //Only fire once
+        }
+
+        public bool IsUsable(Entity user)
+        {
+            if (user is TTTPlayer player)
+            {
+                return true;
+            }
+            return false;
```
mzegar, line 54: Does this work `return user is TTTPlayer`?

mzegar (commented):
It might be good to split up these PRs in the future, just to be separate if the feature is separate. 

But regardless, great work, huge amount of features added.

code/items/equipments/entities/C4Entity.cs
Fires1: Added a comment with some detail.

Considering dropping the [Net] as I believe the assignment of this is only used on the server.

code/items/equipments/entities/C4Entity.cs
Fires1: Good call.

code/items/equipments/HealthkitEquipment.cs
Fires1: How do you feel about the following?

Decoy.cs
```
            using (Prediction.Off())
            {
                if (Input.Pressed(InputButton.Attack1))
                {
                    (owner.Inventory as Inventory).DropEntity(this, typeof(DecoyEntity));
                }
            }
```

Inventory.cs
```
        public bool DropEntity(Entity self, Type entity)
        {
            Entity droppedEntity = Utils.GetObjectByType(entity);
            droppedEntity.Position = Owner.EyePos + Owner.EyeRot.Forward * DropPositionOffset;
            droppedEntity.Rotation = Owner.EyeRot;
            droppedEntity.Velocity = Owner.EyeRot.Forward * DropVelocity;
            return Remove(self);
        }
```
DropPositionOffset and DropVelocity are set at the top.

mzegar: Yeah I think this is fine, might be good to just do the initial inventory cast above for readability.

```csharp
Inventory inventory = owner.Inventory as Inventory
inventory.DropEntity(...)
```

PR #187, opened 2021-08-13
UI: Added `Tabs` UI component

Alf21:
image

### DIFF
--- a/code/ui/components/tabs/Tab.cs
+++ b/code/ui/components/tabs/Tab.cs
@@ -1,0 +1,34 @@
+using System;
+
+using Sandbox.UI;
+using Sandbox.UI.Construct;
+
+namespace TTTReborn.UI
+{
+    public partial class Tab : TTTPanel
+    {
+        public readonly Label TitleLabel;
+        public readonly Tabs Tabs;
+
+        public Action OnSelectTab { get; set; }
+        public Action<Panel> Content { get; set; }
+
+        public Tab(Panel parent, Tabs tabs) : base(parent)
+        {
+            Parent = parent;
+            Tabs = tabs;
+
+            TitleLabel = Add.Label("", "title");
+        }
+
+        public void SetTitle(string title)
+        {
+            TitleLabel.Text = title;
+        }
+
+        protected override void OnClick(MousePanelEvent e)
+        {
+            Tabs.OnSelectTab(this);
+        }
+    }
+}
--- a/code/ui/components/tabs/Tab.cs
+++ b/code/ui/components/tabs/Tab.cs
@@ -8,27 +8,25 @@ namespace TTTReborn.UI
     public partial class Tab : TTTPanel
     {
         public readonly Label TitleLabel;
-        public readonly Tabs Tabs;
 
         public Action OnSelectTab { get; set; }
         public Action<PanelContent> CreateContent { get; set; }
 
-        public Tab(Panel parent, Tabs tabs) : base(parent)
+        private readonly Tabs _parentTabs;
+
+        public Tab(Panel parent, Tabs parentTabs, string title, Action<PanelContent> createContent, Action onSelectTab = null) : base(parent)
         {
             Parent = parent;
-            Tabs = tabs;
+            _parentTabs = parentTabs;
+            CreateContent = createContent;
+            OnSelectTab = onSelectTab;
 
-            TitleLabel = Add.Label("", "title");
-        }
-
-        public void SetTitle(string title)
-        {
-            TitleLabel.Text = title;
+            TitleLabel = Add.Label(title, "title");
         }
 
         protected override void OnClick(MousePanelEvent e)
         {
-            Tabs.OnSelectTab(this);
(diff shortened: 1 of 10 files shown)


code/ui/components/tabs/Tab.cs
```diff
@@ -0,0 +1,34 @@
...
+            Parent = parent;
+            Tabs = tabs;
+
+            TitleLabel = Add.Label("", "title");
+        }
+
+        public void SetTitle(string title)
+        {
+            TitleLabel.Text = title;
+        }
```
mzegar, line 27: Can this just be passed in through the constructor?

code/ui/components/tabs/Tab.cs
```diff
@@ -0,0 +1,34 @@
...
+
+namespace TTTReborn.UI
+{
+    public partial class Tab : TTTPanel
+    {
+        public readonly Label TitleLabel;
+        public readonly Tabs Tabs;
+
+        public Action OnSelectTab { get; set; }
+        public Action<PanelContent> CreateContent { get; set; }
```
mzegar, line 14: Can this just be passed in through the constructor?

code/ui/components/tabs/Tabs.cs
```diff
@@ -0,0 +1,73 @@
...
+        {
+            StyleSheet.Load("/ui/components/tabs/Tabs.scss");
+
+            Header = Add.Panel("header");
+
+            PanelContent = new PanelContent(this);
+            PanelContent.AddClass("content");
+        }
+
+        public Tab AddTab(string title, Action<PanelContent> createContent, Action onSelectTab = null)
```
mzegar, line 27: Similar to the `string` stuff, `Action onSelectTab` should just be an empty action instead of null. Then we don't ever need to null check in the future.

code/ui/components/tabs/Tabs.cs
```diff
@@ -0,0 +1,73 @@
...
+                return;
+            }
+
+            SelectedTab?.SetClass("selected", false);
+
+            SelectedTab = tab;
+
+            SelectedTab.SetClass("selected", true);
+            PanelContent.SetPanelContent(SelectedTab.CreateContent, SelectedTab.TitleLabel.Text);
+            SelectedTab.OnSelectTab?.Invoke();
```
mzegar, line 57: Wouldn't need a null check if we decide with changing the default param above.

code/ui/tttpanel/TTTPanel.cs
```diff
@@ -14,13 +14,13 @@ public bool IsShowing
                 SetClass("hide", !_isShowing);
             }
         }
-        private bool _isShowing = false;
+        private bool _isShowing = true;
```
mzegar, line 17: I'm assuming you checked, but we have no elements that are now showing that used to not show, right? I think we should be covered.

code/ui/components/tabs/Tab.cs
```diff
@@ -0,0 +1,34 @@
...
+
+using Sandbox.UI;
+using Sandbox.UI.Construct;
+
+namespace TTTReborn.UI
+{
+    public partial class Tab : TTTPanel
+    {
+        public readonly Label TitleLabel;
+        public readonly Tabs Tabs;
```
mzegar, line 11: Should these params be `public`? Would there be cases where we need this data available?

mzegar: I'm wondering if there should be a better name for `Tabs` here. Maybe something to indicate that `Tabs` has ownership of `Tab`?

code/ui/components/tabs/Tabs.cs
Alf21: I'm against not using null here. We never know, whether a function is empty then, resetting it would be horrible as well. Additionally, how would you set the default empty action as a param (global empty function shouldn't be the solution). Imo, checking null is better than running a useless empty function.

Alf21 (commented):
Resolved the other requests.

code/ui/components/tabs/Tabs.cs
mzegar: nvm, but could still technically do something like this
https://stackoverflow.com/a/17337403

mzegar (approved):
LGTM, nice stuff, I might take a look at myself changing up how we hand the `_parentTabs` inside of `Tabs`.

There is probably a cleaner way to handle this using something similar to https://api.flutter.dev/flutter/material/TabController-class.html

It also might be worthful to have a vote on whether or not we want to be doing this default `null` param stuff, because we handle it differently in different parts of the codebase. I obviously have one way I prefer, but I rather we stick to one convention than have multiple interfaces that end up behaving differently. I think I remember reading up some `Color` PR that recently went through a similar thing, but right when we hit the function, if it was null, we assigned it to a default value.

code/ui/components/tabs/Tabs.cs
Alf21: no reason doing so because it's optional to  define that action.

PR #189, opened 2021-08-14
Core: Added SettingsSystem

Alf21:
https://user-images.githubusercontent.com/11843790/129463004-27958efd-5599-428c-a0a5-c06d952d8653.mp4

Added language and sprint to settings (saved).

TODO:
- [x] Option to save current settings
- [x] Option to load custom settings (hint whether old settings should be saved as well, otherwise they're getting overwritten)

### DIFF
--- a/code/lang/TTTLanguage.cs
+++ b/code/lang/TTTLanguage.cs
@@ -81,3 +81,35 @@ namespace TTTReborn.Globalization
         }
     }
 }
+
+namespace TTTReborn.Player
+{
+    using Globalization;
+
+    public partial class TTTPlayer
+    {
+        [ClientCmd("ttt_language")]
+        public static void ChangeLanguage(string name = null)
+        {
+            if (name is null)
+            {
+                Log.Info($"Your current language is set to '{TTTLanguage.ActiveLanguage.Data.Name}' ('{TTTLanguage.ActiveLanguage.Data.Code}').");
+
+                return;
+            }
+
+            Language language = TTTLanguage.GetLanguageByCode(name);
+
+            if (language is null)
+            {
+                Log.Warning($"Language '{name}' does not exist. Please enter an ISO (tag) code (http://www.lingoes.net/en/translator/langcode.htm).");
+
+                return;
+            }
+
+            Log.Warning($"You set your language to '{language.Data.Name}'.");
+
+            Settings.ClientSettings.Instance.Language = language.Data.Code;
+        }
+    }
+}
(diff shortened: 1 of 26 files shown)


code/lang/TTTLanguage.cs
```diff
@@ -62,5 +69,47 @@ public static Language GetLanguageByName(string name)
...
+    {
+        [ClientCmd("ttt_language")]
+        public static void ChangeLanguage(string name = null)
+        {
+            if (name is null)
+            {
+                Log.Info($"Your current language is set to '{TTTLanguage.ActiveLanguage.Data.Name}' ('{TTTLanguage.ActiveLanguage.Data.Code}').");
+
+                return;
+            }
```
mzegar, line 99: What's the purpose of providing a default arg of `null` here, if someone calls the function with no args they just get a print statement? Seems a bit off since the name of the function is `ChangeLanguage`

code/player/TTTPlayer.Commands.Server.cs
```diff
@@ -79,6 +79,11 @@ public static void RespawnPlayer(int id)
         [ServerCmd(Name = "ttt_requestitem")]
         public static void RequestItem(string itemName)
         {
+            if (itemName == null)
```
mzegar, line 82: I saw you used `is null` earlier, we should probably just stick to one version.

code/lang/TTTLanguage.cs
Alf21: Yep, to use it like a convar

code/ui/generalhud/menu/Menu.cs
```diff
@@ -52,27 +53,86 @@ public void OpenSettings(PanelContent menuContent)
...
+                            dialogBox.Close();
+                        };
+
+                        FindRootPanel().AddChild(dialogBox);
+
+                        dialogBox.Display();
+                    };
+
+                    fileSelection.Display();
+                });
```
mzegar, line 122: I wonder if we could either condense some of this logic or split it up to make it a bit more readable.

Alf21: Already modified in the other PR, but not as much condensed

code/player/TTTPlayer.Commands.Server.cs
Alf21: Should be done in another refactoring, after we got to a final decision.

mzegar (approved):
LGTM

PR #193, opened 2021-08-15
Make the pistol use pistol ammo

Fires1:


Alf21:
We used Unlimited ammo to avoid issues of not being able to kill someone because of missing crowbaw / weapon spawns.

mzegar:
Yeah I think seems reasonable to keep it unlimited meanwhile.

mzegar:
Closed, we can add this in once we have proper ammo & weapon drops.

PR #194, opened 2021-08-17
FEA: Override inventory property on player

mzegar:


code/player/Inventory.Ammo.cs
```diff
@@ -9,12 +9,12 @@ namespace TTTReborn.Player
 {
     public partial class AmmoInventory
     {
-        private Dictionary<string, int> AmmoList { get; set; } = new();
-        private Inventory Inventory;
+        private Dictionary<string, int> AmmoList { get; } = new();
+        private readonly TTTPlayer _owner;
```
Alf21, line 13: This can be a public var imo, should never change and in that way it could be used at other positions as well if needed :)

code/player/Inventory.Perks.cs
```diff
@@ -8,12 +8,12 @@ namespace TTTReborn.Player
 {
     public partial class PerksInventory
     {
-        private List<TTTPerk> PerkList { get; set; } = new();
-        private Inventory Inventory;
+        private List<TTTPerk> PerkList { get; } = new();
+        private readonly TTTPlayer _owner;
```
Alf21, line 12: Same as above.

code/player/TTTPlayer.cs
```diff
@@ -24,6 +24,12 @@ public partial class TTTPlayer : Sandbox.Player
 
         public bool IsInitialSpawning { get; set; } = false;
 
+        public new Inventory Inventory
+        {
+            get => (Inventory) base.Inventory;
+            private init => base.Inventory = value;
```
Alf21, line 30: Nice trick!

Alf21 (approved):
Beautiful simplifications ❤️

code/player/Inventory.Ammo.cs
mzegar: Two notes...

1. I think we should avoid changing access modifiers for use cases that we don't fully know yet. Whenever access to this is needed, someone can just change it to public.
2. The object relationship within this file has a bit of a code smell. It seems a bit odd that `Inventory` would contain a reference to the `player`, since `player` also has a reference to the base `Inventory`. Another reason why I'd leave it private, this should probably be refactored at some point.

PR #197, opened 2021-08-19
Fix: Scoreboard cut off when player count is high

mzegar:


code/ui/generalhud/scoreboard/Scoreboard.cs
```diff
@@ -70,13 +70,11 @@ private void OnPlayerSpawned(TTTPlayer player)
         private void AddPlayer(PlayerScore.Entry entry)
         {
             ScoreboardGroup scoreboardGroup = GetScoreboardGroup(entry);
-            ScoreboardEntry scoreboardEntry = scoreboardGroup.AddEntry(entry);
-
             scoreboardGroup.GroupMembers++;
+            scoreboardGroup.UpdateLabel();
 
-            _entries.Add(entry.Id, scoreboardEntry);
+            _entries.Add(entry.Id, scoreboardGroup.AddEntry(entry));
```
mzegar, line 76: Just reordering for readability, I did some testing and it seemed like it had no issues.

code/ui/generalhud/scoreboard/Scoreboard.scss
```diff
@@ -14,6 +12,8 @@ Scoreboard {
     display: flex;
     flex-direction: column;
     background-color: #000c;
+    align-self: center;
```
Alf21, line 15: Is this supported?

Alf21 (approved):
I think it is caused by the transform / matrix issue I experienced and reported as well.

code/ui/generalhud/scoreboard/Scoreboard.scss
mzegar: ye, worked within game.

PR #198, opened 2021-08-21
Magneto stick: Rope fix

AMastryukov:
- Fix rope behavior (no longer has weird behavior when hanging body by hands and feet)
- Add ragdoll collisions (between one another)
- Adjust magneto stick pickup behavior (attempt to make it more smooth)


Alf21 (approved):
Nice, finally! ❤️

PR #206, opened 2021-08-24
Newton launcher + crosshair support for chargeable weapons

Krzyska-Filip:
Newton launcher:
I've made working newton launcher with icon.
I haven't made or found model and sound effects yet.

Charge Bar:
To implement it to a weapon you just have to override CreateHudElements() and pass a new argument to Crosshair.Properties (float chargeTime). The only issue might be implementation of custom crosshair mentioned in comments.

I'm not gonna lie the code or approach **might not be** ideal. I tried 15 different ways to implement it but every one of them was much worse than this one.

image


code/items/weapons/NewtonLauncher.cs
```diff
@@ -0,0 +1,135 @@
...
+                        .WithAttacker(Owner)
+                        .WithWeapon(this);
+
+                    tr.Entity.TakeDamage(damageInfo);
+
+                    Vector3 pushDirection = new Vector3(-tr.Normal.x, -tr.Normal.y, tr.Normal.z < 0.5F ? 0.5F : -0.5F);
+                    float chargePercentage = MathX.Clamp((Time.Now - ChargingStartTime) / NEWTON_CHARGE_TIME, 0, 1);
+                    float chargeForce = ((NEWTON_FORCE_MAX - NEWTON_FORCE_MIN) * chargePercentage) + NEWTON_FORCE_MIN;
+
+                    tr.Entity.GroundEntity = null;
```
Alf21, line 96: Why are you doing this?

code/items/weapons/NewtonLauncher.cs
```diff
@@ -0,0 +1,135 @@
...
+                using (Prediction.Off())
+                {
+                    DamageInfo damageInfo = DamageInfo.FromBullet(tr.EndPos, forward * 100 * force, damage)
+                        .UsingTraceResult(tr)
+                        .WithAttacker(Owner)
+                        .WithWeapon(this);
+
+                    tr.Entity.TakeDamage(damageInfo);
+
+                    Vector3 pushDirection = new Vector3(-tr.Normal.x, -tr.Normal.y, tr.Normal.z < 0.5F ? 0.5F : -0.5F);
```
Alf21, line 92: Could you explain the logic at the end?

code/items/weapons/NewtonLauncher.cs
```diff
@@ -0,0 +1,135 @@
...
+                    float chargePercentage = MathX.Clamp((Time.Now - ChargingStartTime) / NEWTON_CHARGE_TIME, 0, 1);
+                    float chargeForce = ((NEWTON_FORCE_MAX - NEWTON_FORCE_MIN) * chargePercentage) + NEWTON_FORCE_MIN;
+
+                    tr.Entity.GroundEntity = null;
+                    tr.Entity.LocalVelocity = pushDirection * chargeForce;
```
Alf21, line 97: Do you know whether that's the way to apply velocity to an object? :o

code/items/weapons/NewtonLauncher.cs
```diff
@@ -0,0 +1,135 @@
...
+                }
+            }
+        }
+
+        public override void CreateHudElements()
+        {
+            if (Local.Hud == null)
+            {
+                return;
+            }
```
Alf21, line 107: This should definitely be avoided, why did you added this check?

code/ui/Crosshair.cs
```diff
@@ -117,6 +142,26 @@ public override void Tick()
...
+
+            if (ChargeTime > 0)
+            {
+                if (Input.Down(InputButton.Attack1))
+                {
+                    if (ChargeStartTime == 0)
+                    {
+                        ChargeStartTime = Time.Now;
+                    }
+                    float chargePercentage = MathX.Clamp((Time.Now - ChargeStartTime) / ChargeTime, 0, 1) * 100;
```
Alf21, line 154: Why using `MathX` here?

Alf21: Could be used inline as well

code/ui/Crosshair.scss
```diff
@@ -1,10 +1,13 @@
-﻿Crosshair {
+Crosshair {
```
Alf21, line 1: What did you changed here? o.O

code/items/weapons/NewtonLauncher.cs
```diff
@@ -0,0 +1,135 @@
...
+        {
+            Vector3 forward = Owner.EyeRot.Forward;
+            forward += (Vector3.Random + Vector3.Random + Vector3.Random + Vector3.Random) * spread * 0.25f;
+            forward = forward.Normal;
+
+            foreach (TraceResult tr in TraceBullet(Owner.EyePos, Owner.EyePos + forward * 5000, bulletSize))
+            {
+                tr.Surface.DoBulletImpact(tr);
+
+                if (!tr.Entity.IsValid() || tr.Entity.IsWorld) return;
```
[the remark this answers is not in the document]
Krzyska-Filip: Yes. Without IsValid it will cause "Object reference not set to an instance of an object." and without IsWorld you are able to push whole map.

code/ui/Crosshair.scss
Krzyska-Filip: Visual Studio did auto code cleaning.

code/items/weapons/NewtonLauncher.cs
Krzyska-Filip: First one calculates time that button was pressed and than it's divided to get normalized result from 0 to 1 but if someone hold button for longer time it will exceed 1 and we don't want a weapon to fire at 110% power.

Second one just get force from min to max based on chargePercentage (previous calculation)

code/items/weapons/NewtonLauncher.cs
Krzyska-Filip: Terry is a little bit different model than classic terrorist from gmod. If you hit him (for example) in upper part of a hand (or forehead) you expect him to get launched  upwards but tr.Noraml.z can return value (negative number) that would push him to the ground instead. It just reduces this threshold where hitting him will push him in to the ground, while still giving you ability to push him down. It also prevents from Terry's space program (launching him way to high).

It might be simplified to only 0.5F but I thought it will look strange if you hit someone from above and still launching him upwards.

code/items/weapons/NewtonLauncher.cs
Krzyska-Filip: It is a way to apply velocity in gmod. I think in sbox it does the same.

code/items/weapons/NewtonLauncher.cs
Krzyska-Filip: It was check from TTTWeapon class and since I override CreateHudElements() I just copied it from parent.
https://github.com/TTTReborn/tttreborn/blob/a919b7f30a44f04b306af418e861119a9bc571f5/code/items/weapons/TTTWeapon.cs#L341

code/ui/Crosshair.cs
Krzyska-Filip: I've explained it above. It can Exceed 1 (100%).

code/items/weapons/NewtonLauncher.cs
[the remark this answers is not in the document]
Alf21: No i meant, don't you wanna use `continue` like below?

code/items/weapons/NewtonLauncher.cs
Alf21: Oh github displaying, I meant the last line `tr.Entity.GroundEntity = null;` :)

code/items/weapons/NewtonLauncher.cs
Alf21: Ah yea, gonna try this out as these "magical numbers" seems to be confusing :D But good solution I think.

code/items/weapons/NewtonLauncher.cs
Alf21: Yea just asking if there is another function that maybe adds some further calculations etc as setting the local velocity gonna kill the already applied velocity

code/items/weapons/NewtonLauncher.cs
Alf21: Okay, that's definitely weird imo :D

code/items/weapons/NewtonLauncher.cs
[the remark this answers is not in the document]
Krzyska-Filip: You are right. I forgot that this check is inside a loop.

code/ui/Crosshair.cs
Alf21: No I meant why using `MathX` instead of `Math` here?
Second comment means instead of defining a single used var, you can put this directly into the function as a param.

code/items/weapons/NewtonLauncher.cs
Krzyska-Filip: I don't think so but there is simple way to fix it.
`tr.Entity.LocalVelocity = tr.Entity.Velocity + pushDirection * chargeForce;`

code/ui/Crosshair.cs
Krzyska-Filip: -MathX is just implemented into sbox.
https://wiki.facepunch.com/sbox/Sandbox.MathX
-I merged this variable with function.

code/items/weapons/NewtonLauncher.cs
Krzyska-Filip: The game has to assume that player isn't standing on the ground in order to launch player. Without it player has to be in air to be launched by newton.

code/ui/Crosshair.cs
Alf21: Yea, I know, but we often run into trouble with sbox-recreated functions that are not as good as cs-sided implementations like the Vector3 system. That's why I suggested usint System's Math.Clamp <https://docs.microsoft.com/en-us/dotnet/api/system.math.clamp?view=net-5.0>, as used in other places in this project as well :)

code/items/weapons/NewtonLauncher.cs
Krzyska-Filip: I found better way to do this by changing tr.Normal to tr.Direction.
`Vector3 pushDirection = new Vector3(tr.Direction.x, tr.Direction.y, Math.Min(tr.Direction.z + 0.5F, 0.75F));`

PR #211, opened 2021-08-27
Add slider component

Fires1:


code/ui/components/slider/Slider.cs
```diff
@@ -0,0 +1,182 @@
...
+namespace TTTReborn.UI
+{
+    /// <summary>
+    /// A horizontal slider. Can be float or whole number.
+    /// </summary>
+    public class Slider : TTTPanel
+    {
+        public Panel Track { get; protected set; }
+        public Panel TrackInner { get; protected set; }
+        public Panel Thumb { get; protected set; }
```
mzegar, line 17: Should these be public? What's the use case for accessing them outside. Changing the style?

code/ui/components/slider/Slider.cs
```diff
@@ -0,0 +1,182 @@
...
+        public float MinValue { get; set; } = 0;
+
+        /// <summary>
+        /// If set to 1, value will be rounded to 1's
+        /// If set to 10, value will be rounded to 10's
+        /// If set to 0.1, value will be rounded to 0.1's
+        /// </summary>
+        public float Step { get; set; } = 1.0f;
+
+        public Slider()
```
mzegar, line 36: Do we want to make this private/protected? Would we want users to be able to use this constructor?

mzegar (commented):
Also a few `var` that will need to be converted.

code/ui/components/slider/Slider.cs
Fires1: Retrieving data off of them.

code/ui/components/slider/Slider.cs
Fires1: I say leave it.

mzegar (approved):

PR #213, opened 2021-08-29
Add entity hint system

Fires1:
Got a nice one here.

This replaces the existing Nameplate system and player corpse popup texts with a new, more usable system.

My primary goal with this addition was to enable hinting to be used in custom entities. I mentioned on Discord that when I implemented the health station I would have loved to have had this, but I wanted to implement it separately so it can be used anywhere. 

Usage is really simple, your *entity* implements a new interface, `IEntityHint`, which enables it to be seen. All UI code currently runs under `TTTPlayer.Hints`, so it's all managed appropriately for each client. `IEntityHint` has two items that are implemented into classes, `bool CanHint(TTTPlayer client)` and `TTTPanel DisplayHint(TTTPlayer client)`, similar to how `IUse` operates. `CanHint` allows you to verify the hint should be shown, such as the `PlayerCorpse` returns true if the body is not identified. `DisplayHint` allows the implementing entity to create some version of a `TTTPanel` to display. Both provide the activating client to the entity.

For ease of use, I have created `EntityHintPanel`, which is a modified version of the old nameplate, with a title and subtitle for two lines of text. In my testing, these resized properly to a variety of lengths with ease. `EntityHintPanel` also exposes `Action<EntityHintPanel> DoTick` to the implementing entity. This can be seen in the nameplate's implementation in which it runs code to update the element if the player's health changes. As well as a method for modifying the base panel's style, and another method to including the default nameplate background. These methods are implemented in the `WithXYZ` style, which returns an instance of itself, allowing the panels to be created inline and with ease.

Overall, this system is beneficial as it removes the need to manually implement hinting systems, as well as the customizability allows any sort of panel to be shown to the player when needed. Due to the nature of the panels, I recommend that these panels do not implement any sort of functionality besides displaying information.

Here's the pretty pictures and gifs...

Identical nameplate implemented with the entity hint system

Confirmation hint implemented with the entity hint system

[CLICK TO SEE GIF: Works with no background or with background on any sort of entity](https://giant.gfycat.com/BareHealthyArmyant.mp4)

[Also works with weapons!](https://gfycat.com/thunderousdearbabirusa)


code/player/PlayerCorpse.cs
```diff
@@ -119,5 +121,17 @@ public ConfirmationData GetConfirmationData()
                 Suicide = Suicide
             };
         }
+
+        public bool CanHint(TTTPlayer client) => !IsIdentified;
```
mzegar, line 125: `ShouldShowHint` `IsShowingHint` `ShowHint`

code/player/PlayerCorpse.cs
```diff
@@ -119,5 +121,17 @@ public ConfirmationData GetConfirmationData()
...
+
+        public TTTPanel DisplayHint(TTTPlayer client)
```
mzegar, line 127: `Hint` `HintToDisplay`

Fires1:
- [ ] Add "HintDistance" to `IEntityHint` so entities can choose when the hint should display, rather than a hard coded number.

mzegar (commented):
I don't have much, really nice changes, I got suggestions potentially for renames, but they are completely up to you. I'm fine with either.

mzegar (approved):
I'm good with this going in without the distance change for now, can add an issue for it.

Issue #204, opened 2021-08-22
Damage Overlay bugged


Alf21:
## Prerequisites

- [x] I have searched [existing bug reports](https://github.com/TTTReborn/ttt-reborn/issues?q=label%3Aissue%2Fbug-report) and confirmed my issue is not already reported.

## Summary
Some s&box update broke the damage overlay. Opacity with background image interaction seems to be bugged.

## Steps to reproduce
Just receive damage in-game. The damage overlay should look like sh*t.

mzegar:
@Krzyska-Filip do you want to throw up a PR for this? Thanks for looking into this.

GravityWolfNotAmused:
Can this be closed?

Krzyska-Filip:
(after my PR) 
Damage indicator had minor bug where executing suicide while being dead would display damage indicator. Alternative was that last dmg before death was not displaying properly.

(after recent AliveHud PR / I haven't checked it)
Damage indicator will dissapear after death which will look a bit odd (a least for me)

(Ideal behaviour)
Damage indicator display after death but does not trigger when player is dead. Possible, it would mean that we have to move damage indicator from AliveHUD to GeneralHUD and implement PlayerState.Dying unless someone know better way to solve it.

mzegar:
Yeah that should fix the issue, let's keep this GitHub issue for this as it's tangentially related.

PR #214, opened 2021-08-29
Fix for #204 (Damage Overlay bugged)

Krzyska-Filip:
Fixed issue where maximum duration
of indicator was multiplied instead of getting max value out of it.

(Example)
If player types kill in console he would receive 1000 damage which will cause overlay to stay for 100s.

### DIFF
--- a/code/ui/Hud.cs
+++ b/code/ui/Hud.cs
@@ -1,6 +1,3 @@
-using System;
-using System.Collections.Generic;
-
 using Sandbox;
 using Sandbox.UI;
 
--- a/code/ui/Hud.cs
+++ b/code/ui/Hud.cs
@@ -1,3 +1,6 @@
+using System;
+using System.Collections.Generic;
+
 using Sandbox;
 using Sandbox.UI;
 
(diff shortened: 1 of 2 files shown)


Krzyska-Filip:
After further testing I found another issue where after player spawn/re-spawn after each round will cause one more damage indicator to appear. 5 deaths = 5 new indicators working at the same time.

(Example)
0 respawns:
image

5 respawns:
image

Krzyska-Filip:
Found another Issue with events, related to previous one. Same result different approach. On hot reload event are called more times. 5 reloads = 5 events called.

(Example)
0 reloads:
image

5 reloads:
image

Krzyska-Filip:
Map triggers that deal damage per tick increase time that indicator is shown properly but if you get out of trigger, it for some reason set remaining time to value = 0.033333335. I don't know the reason behind this "magic". I've been debugging 10 lines of code for 5 hours and I still don't have any idea what might me wrong.

code/ui/Hud.cs
```diff
@@ -1,3 +1,6 @@
+using System;
+using System.Collections.Generic;
+
```
Fires1, line 3: Still need these?

Fires1 (approved):

Fires1:
Looks to be all good and clean changes. Appreciate the contribution.

Ignore me fighting with Github above to label/assign/review properly haha.

mzegar:
Hi, this might be a separate bug, but I think it would be easy to fix and semi-related to these changes. You can trigger the damage overlay by pressing the suicide button even as a spectator.

Krzyska-Filip:
The best and cleanest solution to this is adding `player.LifeState == LifeState.Dying` but currently it is not implemented in TTTR.
In my opinion we should return to that later instead of creating a lot of bools. Alternatively we can leave it for alive player but then last damage before kill won't be displayed.

mzegar:
@Krzyska-Filip np!

PR #217, opened 2021-09-11
Shop: Added `Shop` type, server to client shop syncing and shop settings

Alf21:
TODO
- [x] Shop types
- [x] Shop server to client syncing
- [x] Shop settings
- [ ] Shop editor
- [ ] Random shops

### DIFF
--- a/code/items/IBuyableItem.cs
+++ b/code/items/IBuyableItem.cs
@@ -6,9 +6,9 @@ namespace TTTReborn.Items
 {
     public struct ShopItemData
     {
-        public string Name;
+        public string Name { get; set; }
         public string Description;
-        public int Price;
+        public int Price { get; set; }
         public SlotType SlotType;
         public Type Type;
 
--- a/code/items/IBuyableItem.cs
+++ b/code/items/IBuyableItem.cs
@@ -4,13 +4,18 @@ using TTTReborn.Player;
 
 namespace TTTReborn.Items
 {
-    public struct ShopItemData
+    public class ShopItemData
     {
         public string Name { get; set; }
-        public string Description;
-        public int Price { get; set; }
-        public SlotType SlotType;
-        public Type Type;
+        public string Description = "";
+        public int Price { get; set; } = 0;
+        public SlotType? SlotType = null;
+        public Type Type = null;
+
+        public ShopItemData(string name)
+        {
+            Name = name;
+        }
 
         public bool IsBuyable(TTTPlayer player)
         {
@@ -21,7 +26,7 @@ namespace TTTReborn.Items
 
             if (Type.IsSubclassOf(typeof(TTTEquipment)) || Type.IsSubclassOf(typeof(TTTWeapon)))
             {
-                return !player.Inventory.IsCarryingType(Type) && player.Inventory.HasEmptySlot(SlotType);
+                return !player.Inventory.IsCarryingType(Type) && player.Inventory.HasEmptySlot((SlotType) SlotType);
             }
 
             return false;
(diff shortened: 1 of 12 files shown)


code/items/IBuyableItem.cs
```diff
@@ -4,13 +4,18 @@
...
-    public struct ShopItemData
+    public class ShopItemData
     {
-        public string Name;
-        public string Description;
-        public int Price;
-        public SlotType SlotType;
-        public Type Type;
+        public string Name { get; set; }
+        public string Description = "";
```
mzegar, line 10: Any reason why one is property and the other one isn't?

code/player/TTTPlayer.Nameplate.cs
```diff
@@ -54,28 +54,32 @@ public TTTPanel DisplayHint(TTTPlayer client)
...
-                       //Set health group related data on bottom label.
-                       float health = Health / MaxHealth * 100;
-                       HealthGroup healthGroup = GetHealthGroup(health);
+                {
+                    if (this == null || !IsValid)
+                    {
+                        return;
+                    }
+
+                    if (Health == 0 && LifeState == LifeState.Alive)
```
mzegar, line 63: Do we just want to make a call to `IsSpectator` here?

code/roles/TTTRole.cs
```diff
@@ -50,7 +63,18 @@ public virtual void OnDeselect(TTTPlayer player)
 
         }
 
-        public virtual bool CanBuy() => false;
+        // serverside function
```
mzegar, line 66: Could put an `Assert(Host.IsServer)` if you want to make sure whenever someone calls this it's on the server.

code/ui/alivehud/quickshop/QuickShop.cs
```diff
@@ -35,6 +35,17 @@ public void Update()
             _content.Update();
         }
 
+        [Event("tttreborn.shop.change")]
+        public static void ShopChanged()
```
mzegar, line 39: minor nit, maybe `OnShopChanged`

code/items/IBuyableItem.cs
Alf21: Just Name and Price should be included into json. That's why. Otherwise, I had to do { internal get; internal set } which wouldn't make any difference

code/player/TTTPlayer.Nameplate.cs
Alf21: dk, just fixed a bug that's bugged the game while testing... Would it keep as it is until a rework is planned (or in Alex's branch)

code/roles/TTTRole.cs
Alf21: Could you provide an example?

mzegar: to the first line of `InitShop()` you can write...

`Assert(Host.IsServer)` and it'll throw an exception if `IsServer` is false.

Alf21: Did you tried this? As `Assert` is no function available in `Sandbox` namespace...

code/player/TTTPlayer.Shop.cs
```diff
@@ -10,13 +15,85 @@ public enum BuyError
         None,
         InventoryBlocked,
         NotEnoughCredits,
-        RoleRestriction
+        NoAccess
+    }
+
+    public class Shop
+    {
+        public List<ShopItemData> Items { set; get; } = new();
```
mzegar, line 23: Can this just be initialized in the constructor?

code/roles/TTTRole.cs
mzegar: `Host.AssertServer()` mb

mzegar (approved):
LGTM, feel free to get another person to review as well.

code/player/TTTPlayer.Shop.cs
Alf21: Why? We never initializing in the constructor :D

PR #222, opened 2021-09-18
Code Clean Up

GravityWolfNotAmused:
As I was looking around the codebase I noticed some inconsistencies in the way some things are written, not logically, but syntactically speaking.

The most common has to be how constructors are called. I have changed all that I found to new() as the variable type can be inferred. The second is between using the using directive or specifying the namespace on the variable type. In some instances this can cause ambiguity; however, the ones removed did not possess this ambiguity.

*Removed question for issue submission*

### DIFF
--- a/code/settings/Settings.cs
+++ b/code/settings/Settings.cs
@@ -113,7 +113,7 @@ namespace TTTReborn.Settings
     {
         public const string SETTINGS_FILE_EXTENSION = ".settings.json";
 
-        public static string GetJSON<T>(T settings, bool compressed = false) where T : Settings
+        public static string GetJSON(T settings, bool compressed = false) where T : Settings
         {
             JsonSerializerOptions options = new()
             {
--- a/code/settings/Settings.cs
+++ b/code/settings/Settings.cs
@@ -100,11 +100,11 @@ namespace TTTReborn.Settings
         {
             if (Host.IsClient)
             {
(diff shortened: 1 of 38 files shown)


AMastryukov (approved):

mzegar (approved):
LGTM, thanks.

GravityWolfNotAmused:
@mzegar You are welcome, I am going to do more prowling, and cleaning up after dinner. 
I made another commit to up one issue with formating that wasn't to the standard.

code/settings/Settings.cs
```diff
@@ -115,17 +115,17 @@ public partial class SettingFunctions
 
         public static string GetJSON<T>(T settings, bool compressed = false) where T : Settings
```
Alf21, line 116:
```suggestion
        public static string GetJSON(T settings, bool compressed = false) where T : Settings
```

Alf21 (commented):
GJ, ty! ❤️

GravityWolfNotAmused:
@Alf21 I have committed your suggestion. I guess it went right past my eyes. 👍

Alf21 (approved):

GravityWolfNotAmused:
@Alf21 There is an issue with the changes you asked for

GravityWolfNotAmused:
@Alf21 We need that generic as we are passing a variable of type T.

Alf21 (approved):

Issue #212, opened 2021-08-27
Crosshair and drowning indicator stay on player death.


Krzyska-Filip:
## Prerequisites

- [x] I have searched [existing bug reports](https://github.com/TTTReborn/ttt-reborn/issues?q=label%3Aissue%2Fbug-report) and confirmed my issue is not already reported.

## Summary
Crosshair stays on player's screen after death that is not suicide.
Same with drowning indicator when you drown.
![image](https://user-images.githubusercontent.com/45741043/131174882-fbb81073-7634-435e-bae3-1e3a1707a4d0.png)

## Expected behavior
Crosshair and drowning indicator should be removed\hidden after player death

## Actual behavior
Both crosshair and drowning indicator are displayed after death.

## Steps to reproduce
To reproduce crosshair bug you have to die while not ending the current round.
To reproduce drowning indicator you have to drown.

GravityWolfNotAmused:
#212
At the moment you are adding the UI elements to the HudEntity parent, so if we set AliveHud's IsEnabled = false we are setting AliveHud and its children to disabled, but we never added children to AliveHud, we add them to the HudEntity's RootPanel.

If we add these panels to AliveHud instead of the parent, we have to redo the UI to be in the correct position which I don't understand fully yet. (sorry)

I foresee that this collection will grow as the game mode does, and so will the time to disable all of the elements. o(n)

With this being said, would you rather collect them all and disable them? or should we redo the UI, so they are in view and add them as a child of AliveHud?

https://github.com/TTTReborn/tttreborn/blob/bd9262f5f9749dde56096db7295066151d8eb24a/code/ui/Hud.cs#L96-L103

mzegar:
Fixed with #212

PR #223, opened 2021-09-18
Bug #212

GravityWolfNotAmused:
## Link to Issue:
#212 

## Issue
Crosshair and Drowning Indicator are not disabled when the player dies.

## Tests

- When a player dies, and the round ends due to time up.
   - **Passed**
- When a player dies, and the round ends due to no Traitors/Innocents.
   - **Passed**

## Expected Results 
- **When the player dies, the crosshair and the drowning indicator do not show to the player.**
- **When the player respawns, the crosshair will show to the player.**


code/ui/Hud.cs
```diff
@@ -51,55 +48,60 @@ private void OnPlayerSpawned(TTTPlayer player)
...
-                Parent.AddChild<PlayerInfoDisplay>();
-                Parent.AddChild<InventoryWrapper>();
-                Parent.AddChild<ChatBox>();
-
-                Parent.AddChild<VoiceChatDisplay>();
-                Parent.AddChild<GameTimerDisplay>();
-
-                Parent.AddChild<VoiceList>();
+                AddClass("fullscreen");
+                AddChild<Crosshair>();
```
mzegar, line 72: Let's throw the crosshair inside of `AliveHud`, since we already disable / enable it based on being alive.

GravityWolfNotAmused: Yes, but I tried this and it fails to draw the crosshair on spawn until I toggle the console on and off.

Is there a sync, or refresh I need to perform?

GravityWolfNotAmused: Even if I switched to doing one of the following it still fails to draw the crosshair:
```csharp
AliveHudPanel.Enabled = value;
AliveHudPanel.GetCrosshair().Enabled = value;
```

```csharp
AliveHudPanel.Enabled = value;
```

mzegar: Hmm, when you open and close the dev console, the cross-hair appears? I've never ran into something like this. It should work like all of the other HUD elements in `AliveHud`.

GravityWolfNotAmused: **After spawning in:**
image

**After toggling console on and off:**
image

GravityWolfNotAmused: I do believe that this is a result of toggling AliveHud on and off, but don't know how I would get around this.

mzegar: I'll hopefully take a look at this sometime and see if I can figure it out.

code/ui/Hud.cs
```diff
@@ -51,55 +48,60 @@ private void OnPlayerSpawned(TTTPlayer player)
...
+            }
 
-                Parent.AddChild<InfoFeed>();
-                Parent.AddChild<InspectMenu>();
-                Parent.AddChild<PostRoundMenu>();
-                Parent.AddChild<Scoreboard>();
-                Parent.AddChild<Menu.Menu>();
+            public Panel GetCrosshair()
+            {
+                return (Panel) GetChild(0);
```
Alf21, line 93: dk whether Crosshair should be the first element nor handled in this way. E.g. the radar points should be before the crosshair.

GravityWolfNotAmused: I am not a fan of how I had to do it either, I just wish that I could just turn on and off the parent, but it's giving me hell when I do it.

GravityWolfNotAmused:
> Hmm, when you open and close the dev console, the cross-hair appears? I've never run into something like this. It should work like all of the other HUD elements in `AliveHud`.

It seems that we wish to investigate the reason why turning off, and on AliveHud isn't making Crosshair visible.

I ran more tests on just disabling AliveHud instead of doing something different for the crosshair as mentioned by @mzegar.

# Just Toggling AliveHud#Enabled

## When I Die
So I added print statements for the enabled variable for different parts of the hud.

AliveHudPanel#Enabled = false,
Crosshair did not have disabled css class,
Crosshair#Enabled = true

image

To my understanding this is correct, it has disabled the AliveHudPanel, but its Children are still enabled and they do not have the disabled css class. Unless sandbox should have turned off all of the disabled CSS classes?

#### Video of Dying
https://user-images.githubusercontent.com/24533882/133914481-2c08b49e-448b-40b1-a819-186df170e38e.mp4

## When I spawn

I am met with no crosshair, but the DrowningIndicator does reactivate when I go underwater; however, the strange thing is that when I toggle on and off the console, my crosshair shows up...

#### Video of Spawning, and fixing Crosshair visibility

https://user-images.githubusercontent.com/24533882/133914497-80eed7b9-418d-4692-9011-3ad8f92540ce.mp4

# Something that works as intended, and is the last commit:

If I create a function inside of our Panel to manually turn on and off the disabled css class for all of the children, and turn then back on when I spawn everything works as intended. I am assuming the other methods I have tried may be an issue with S&ndbox.

```csharp
class Panel : Sandbox.UI.Panel
{
     //...Fill in other code
     
     public void SetChildrenEnabled(bool status)
      {
            Children.ToList().ForEach(child => ((Panel)child).Enabled = status);
      }        
}
```

GravityWolfNotAmused:
Now that I'm in bed, and I'm thinking about it. It's probably best just to do this on Panel's Enabled property's set function. I'll make that change tomorrow. Better that way.

mzegar:
Hmm... I think it'd be good to figure out why the `Enable` / `Disable` is working for everything else but the crosshair. I spent some time and didn't have any luck, but I'll try to further investigate.

GravityWolfNotAmused:
> Hmm... I think it'd be good to figure out why the `Enable` / `Disable` is working for everything else but the crosshair. I spent some time and didn't have any luck, but I'll try to further investigate.

This would be the only other option if we cannot find out a reason for it. 
I am thinking it has to be a sandbox issue, but my lack of experience with sandbox lessens the validity of this statement.

I tried putting this logic inside of the setter for Enabled, but that causes us to set the children, and the children's children...etc. Which isn't favorable.

It failed in my testing, and acted the same way as it originally did with the toggle of the console.

mzegar:
Fixed, maybe it had something to do with the `Crosshair` panel having two center classes applied to them.

mzegar (approved):

GravityWolfNotAmused:
Will have to keep that in mind later if this ever happens again with other parts of the UI.

Alf21 (approved):

Issue #157, opened 2021-07-26
Add idle-check (auto afk system)


Alf21:

GravityWolfNotAmused:
I think i want to attack this, it seems like a simple utility, and would be great for the exposure to the library. I'll begin planning tomorrow :)

PR #227, opened 2021-09-21
AFK System

GravityWolfNotAmused:
**Issue:** #157 

**Original concept:**
- The client keeps track of AFK time, last known rotation, and when a key is pressed to update the time to kick.
- The server is used by server command to kick the caller of the console command from the server.

I am putting this up really quickly not because it's rushed, but I want to hear your opinion on it, so I can improve my knowledge. Please try to refrain from taking the PR over, or changing. <3

Thank you.


code/player/TTTPlayer.AFK.cs
```diff
@@ -0,0 +1,41 @@
...
+            bool pressedAnyKeyPressed = Buttons.Any(button => Input.Pressed(button) || Input.Down(button));
+
+            if (pressedAnyKeyPressed || (LastKnownRotation.HasValue && LastKnownRotation != Rotation))
+            {
+                _timeSinceLastAction = Time.Now;
+                LastKnownRotation = Rotation;
+            }
+
+
+            if (TimeToKick < Time.Now)
```
Krzyska-Filip, line 28: I think it would be better to use `TimeSince` variable.

```suggestion
        private static int SecondsTillKick => ServerSettings.Instance.AFK.MinutesTillKick * 60;
        
        private TimeSince _timeSinceLastAction = 0f;
        private Rotation? LastKnownRotation;

        private void TickAFKSystem()
        {
            bool pressedAnyKeyPressed = Buttons.Any(button => Input.Pressed(button) || Input.Down(button));

            if (pressedAnyKeyPressed || (LastKnownRotation.HasValue && LastKnownRotation != Rotation))
            {
                _timeSinceLastAction = 0;
                LastKnownRotation = Rotation;
            }
            if (SecondsTillKick < _timeSinceLastAction)
```

Krzyska-Filip, line 28: I think it would be better to use `TimeSince` variable.

```suggestion
        private static int SecondsTillKick => ServerSettings.Instance.AFK.MinutesTillKick * 60;
        
        private TimeSince _timeSinceLastAction = 0f;
        private Rotation? LastKnownRotation;

        private void TickAFKSystem()
        {
            bool pressedAnyKeyPressed = Buttons.Any(button => Input.Pressed(button) || Input.Down(button));

            if (pressedAnyKeyPressed || (LastKnownRotation.HasValue && LastKnownRotation != Rotation))
            {
                _timeSinceLastAction = 0;
                LastKnownRotation = Rotation;
            }
            if (SecondsTillKick < _timeSinceLastAction)
```

Fires1:
Should definitely use the edit with the `Time_` structs that S&box has built in since all that counting logic is completely out.

Within our `Game` object we also have an OnSecond tick, which as far as I know, we don't trip to the client, but I think it would be worth it to trip a "OnRoundSecond" on the client side, and run slower checks like this. It's a small irk but it saves slightly on tick time to process stuff like this.

I would also push for this to force players to spectate, rather than kicking, but that would also require a "force stay spectator" mode like in the original TTT. I don't think many would have gripes to bring that feature over, I find it a lot more useful when you suddenly have to go AFK and you end up staying in the server, rather than being completely booted.

Fires1 (changes requested):
Use `TimeSince`.

GravityWolfNotAmused:
@Fires1 I agree I rather force the player into spectating, but I wanted to start with the most simple solution to give myself more experience with different ideas. Please note the end goal is to have it in the server config to force players into spectating or kick them.

GravityWolfNotAmused:
This is still very much a work inprogress.

code/player/TTTPlayer.AFK.cs
GravityWolfNotAmused: I have switched to TimeSince. Thank you.

GravityWolfNotAmused:
At the very moment, I cannot interact with the settings menu without inverting the order of AliveHud, and GeneralHud.
I am harboring the change in my fork because I don't think you all want me to invert it.

Any pointers on the issue, or any other changes in format or coding?

Krzyska-Filip:
You could display warning to a player ~10 seconds before kick/spectate.

Fires1:
> Do we have a standard way to send these notifications?

As far as I know, we don't have a standard "very fucking obvious" notification, like the bottom box that appears at the start of rounds in TTT1 that gives you information about your role. Within my pending PR for map entities, there's a helper method for adding custom feed entries, which would be the best for notifying once merged. I'd say make an issue for it and get it fixed once both this PR and the map entities are merged. TTT1 notified in the confirmation feed with AFK warnings, so it won't feel out of place there.

Fires1 (approved):
LGTM if UI issues are resolved.

code/player/TTTPlayer.AFK.cs
```diff
@@ -8,34 +10,30 @@ namespace TTTReborn.Player
...
-            if (pressedAnyKeyPressed || (LastKnownRotation.HasValue && LastKnownRotation != Rotation))
+            if (pressedAnyKeyPressed || (_lastKnownRotation.HasValue && _lastKnownRotation != Rotation))
             {
-                _timeSinceLastAction = Time.Now;
-                LastKnownRotation = Rotation;
+                _timeSinceLastAction = 0f;
+                _lastKnownRotation = Rotation;
             }
 
+            //Log.Warning($"Time: {_timeSinceLastAction}, Second to wait: {SecondsTillKick}, Bool: {_timeSinceLastAction > SecondsTillKick}");
```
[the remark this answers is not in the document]
GravityWolfNotAmused: Will do.

GravityWolfNotAmused:
> LGTM if UI issues are resolved.

The settings UI can be fixed by changes the order of which GeneralHud, and AliveHud are added; however, last time I tried to change these they were reverted, and the issue ended up being a css class. If there is less fear of the huds causing conflict I will gladly add that change to my changes.

GravityWolfNotAmused:
When I get home I will commit my changes to the Huds, and remove the debugging Warning and we can go into a full review to be sure that there are not any additional requirements from other devs.

Fires1:
I'm not qualified to speak on the new UI changes, since I have not yet had a change to look over them since they got merged so quickly.

Are you saying the settings menu is completely unusable on your branch?

GravityWolfNotAmused:
Correct, without changing the order of GeneralHud and AliveHud.

GravityWolfNotAmused:
image
This is the change that I am harboring because I think a UI dev would rather me not do this, but I am unsure. All I have done is add to the menu, no other changes were done to that menu.

GravityWolfNotAmused:
I have made additional changes; however, I am sure that some of the reorganizations might be something that people can critique a bit.

# Added:
- Changed HasMinimumPlayers to exclude those in force spec.
   - This stops rounds with 1 spec and 1 player from happening
- Check for win condition for when the system moves a player to force spec
   - This is for when there are 2 players left on the server, and one is marked AFK. Without this, the round will wait for the timer.
- Add force spec toggle and general tab to the client settings menu, so I could take myself out of force spectator. (BUG 1)

# Note:
There is one bug I am facing and investigating, but it seems more of an escape bug, or I am just not understanding a better way to do it.

# Bug 1: 
If the user has already opened the settings menu onto the General tab, and closed/hid it if they are put into force spec then the switch status will not reflect the changes.

# Link:
https://github.com/TTTReborn/tttreborn/pull/227/commits/e497e32fe10336552c80c652dcbbbc29db2490c4

GravityWolfNotAmused:
Other than the one bug, and #228 there should be completely functional. If you experience the bug listed above you can just change tabs, or go back home and navigate back to Client > General tab to refresh the data.

GravityWolfNotAmused:
TODO: Change execution of end round checks when to an event instead of casting and calling InProgress round.

GravityWolfNotAmused:
I am finished making changes to this and require input.
I fixed my issues with players having the settings menu hid on the client settings tab.
Please let me know of any changes you want, or do not want to see.

Thanks, everyone.

GravityWolfNotAmused:
Oops...Wrong oringally posted in the wrong place, this PR is on hold until #229 is merged because of possible merge conflicts.

code/gamemode/Game.Settings.cs
```diff
@@ -3,11 +3,20 @@ namespace TTTReborn.Settings
...
+        public Categories.AFK AFK { get; set; } = new Categories.AFK();
     }
 
 
     namespace Categories
     {
+        public partial class AFK
+        {
+            /// <value><c>ShouldKickPlayers</c> is if the system should kick the player versus moving them to spectators.</value>
+            public bool ShouldKickPlayers { get; set; } = true;
```
mzegar, line 15: I think by default original TTT used to just move players to spectator instead of kicking them, I kinda of agree with sticking with this default. Thoughts?

code/global/RPCs.cs
```diff
@@ -113,6 +113,12 @@ public static void ClientConfirmPlayer(TTTPlayer confirmPlayer, PlayerCorpse pla
             }
         }
 
+        [ClientRpc]
+        public static void ClientOnMovedToForceSpectator(bool value)
+        {
+            Hud.Current.AliveHudInstance.Enabled = false;
+        }
```
mzegar, line 120: We need this? I feel like this should be done inside of `MakeSpectator`

code/global/Utils.cs
```diff
@@ -88,7 +103,7 @@ public static IEnumerable<Client> GetClientsSpectatingPlayer(TTTPlayer player)
 
         public static bool HasMinimumPlayers()
         {
-            return Client.All.Count >= Settings.ServerSettings.Instance.Round.MinPlayers;
+            return Client.All.Count - GetForcedSpectatingPlayers().Count >= Settings.ServerSettings.Instance.Round.MinPlayers;
```
mzegar, line 106: I'm not sure how much more performant, but we could just make a `GetNonForcedSpectatingPlayers()` call instead of doing a subtraction here.

code/player/TTTPlayer.AFK.cs
```diff
@@ -0,0 +1,68 @@
...
+using Sandbox;
+
+using TTTReborn.Settings;
+
+namespace TTTReborn.Player
+{
+    public partial class TTTPlayer
+    {
+        public readonly static List<InputButton> Buttons = Enum.GetValues(typeof(InputButton)).Cast<InputButton>().ToList();
+        private static int SecondsTillKick => ServerSettings.Instance.AFK.MinutesTillKick * 60;
```
mzegar, line 14: `SecondsTillKick` and then the server setting is `MinutesTillKick`, we should pick one. I think typically in the codebase we define everything in `Seconds`.

code/player/TTTPlayer.AFK.cs
```diff
@@ -0,0 +1,68 @@
...
+                return;
+            }
+
+            if (IsForcedSpectator || IsSpectator)
+            {
+                _timeSinceLastAction = 0;
+                return;
+            }
+
+            bool pressedAnyKeyPressed = Buttons.Any(button => Input.Pressed(button) || Input.Down(button));
```
mzegar, line 34: I think we can just do either `Pressed` or the `Down` check, no need for both?

code/player/TTTPlayer.AFK.cs
```diff
@@ -0,0 +1,68 @@
...
+
+namespace TTTReborn.Player
+{
+    public partial class TTTPlayer
+    {
+        public readonly static List<InputButton> Buttons = Enum.GetValues(typeof(InputButton)).Cast<InputButton>().ToList();
+        private static int SecondsTillKick => ServerSettings.Instance.AFK.MinutesTillKick * 60;
+
+        private TimeSince _timeSinceLastAction = 0f;
+        private Rotation? _lastKnownRotation;
```
mzegar, line 17: Do we really need to track rotation as well? I think that tracking button presses are enough, or even just tracking rotation could work. Open to hearing thoughts.

code/player/TTTPlayer.Spectating.cs
```diff
@@ -83,5 +84,30 @@ public void MakeSpectator(bool useRagdollCamera = true)
 
             ShowFlashlight(false, false);
         }
+
+        public void ForceSpectator()
```
mzegar, line 88: since we do `IsForcedSpectator = !IsForcedSpectator;` this sounds like a `ToggleForceSpectator` instead of `ForceSpectator`.

code/player/TTTPlayer.Spectating.cs
```diff
@@ -83,5 +84,30 @@ public void MakeSpectator(bool useRagdollCamera = true)
...
+                    {
+                        round.ExecuteOnWinCondition();
+                    }
+
+                    RPCs.ClientOnMovedToForceSpectator(To.Single(this), IsForcedSpectator);
+                }
+            }
+
+            string toggle = IsForcedSpectator ? "activated" : "deactivated";
+            Log.Info($"You {toggle} force spectator.");
```
mzegar, line 110: I don't think we need to log this.

code/rounds/InProgressRound.cs
```diff
@@ -210,5 +198,15 @@ public override void OnSecond()
...
+
+        public void ExecuteOnWinCondition()
+        {
+            TTTTeam result = IsRoundOver();
+
+            if (result != null)
+            {
+                LoadPostRound(result);
+            }
+        }
```
mzegar, line 210: We don't need this, check out `OnSecond()` inside of `InProgressRound()`. Technically, we should probably move to an event-based system where we only check if a player dies, leaves, etc. We can make a ticket for this.

code/ui/generalhud/menu/Menu.Settings.Server.cs
```diff
@@ -22,12 +22,44 @@ internal void CreateServerSettings(PanelContent tabContent, ServerSettings serve
...
+
+                ConsoleSystem.Run("ttt_serversettings_send", SettingFunctions.GetJSON(serverSettings, true));
+            });
         }
 
         private void AddSprintSettings(PanelContent tabContent, ServerSettings serverSettings)
         {
             Sandbox.UI.Panel sprintPanel = tabContent.Add.Panel("sprint");
-            sprintPanel.Add.Label($"Sprint enabled?");
+            sprintPanel.Add.Label($"Sprint enabled?").AddTooltip("Wether or not sprint should be enabled on the server");
```
mzegar, line 62: Spelling mistake

code/gamemode/Game.Settings.cs
GravityWolfNotAmused: Giving them the option to kick over just moving to spectate is a great idea, but I agree that the default should likely be what people are used to in previous versions.

code/global/RPCs.cs
GravityWolfNotAmused: MakeSpectator does set its health and state, but is not what calls "OnKilled", "OnKilled" is what calls MakeSpectator. Also, we cannot manually call this method, which will cause a stack overflow. Do you have a way for me to trigger OnKilled without causing this?

https://github.com/TTTReborn/tttreborn/blob/24eac8ada69f3cf43b96f5688767e9f8ecc76688/code/player/TTTPlayer.Spectating.cs#L74-L85

https://github.com/TTTReborn/tttreborn/blob/24eac8ada69f3cf43b96f5688767e9f8ecc76688/code/rounds/InProgressRound.cs#L24-L37

code/global/Utils.cs
GravityWolfNotAmused: I agree. Will make this change.

code/player/TTTPlayer.AFK.cs
GravityWolfNotAmused: Agreed. Will make this change.

code/player/TTTPlayer.AFK.cs
GravityWolfNotAmused: I rather check for both because the press is only on the action of down and released, but holding the button is not taking into account without Down.

code/player/TTTPlayer.AFK.cs
GravityWolfNotAmused: This is for instances where the player isn't moving but is looking around.

It's not super big to do, or not do. I just figured it was another small value I could check against to see if it was different. Open for a different value, but I think the check should be more than just a button check.

code/player/TTTPlayer.Spectating.cs
GravityWolfNotAmused: Agreed. Will make change

code/player/TTTPlayer.Spectating.cs
GravityWolfNotAmused: Agreed, removed.

code/global/RPCs.cs
mzegar: Hmm... I'm thinking maybe `MakeSpectator` is what calls for the disable/deletion of HUD. There is definitely some way to handle this.

GravityWolfNotAmused: I agree, but don't I have to do this on client side?

mzegar: here are some options, you'd have to give me some time to go digging around for the best one.

1. `MakeSpectator` damages the player for a huge amount, thus killing the player and triggering `OnKill` (which then disables the UI)
2. `MakeSpectator` runs the `ttt_player_onkilled` event which triggers the RPC to disable the hud

GravityWolfNotAmused: I manually triggered OnKill inside of ToggleForcedSpectator, if this were to be written we would have a stack overflow. Seems that we still have the issue with AliveHud even after @Alf21's changes.

code/ui/generalhud/menu/Menu.Settings.Server.cs
GravityWolfNotAmused: Fixed

code/rounds/InProgressRound.cs
GravityWolfNotAmused: I wasn't able to get that to automatically trigger, and I agree the event system sounds better I just failed at being able to make events/hooks.

code/player/TTTPlayer.AFK.cs
mzegar: I'm just obsessing over performance that probably won't really matter, but I guess my idea is that we don't need an overly complex anti-AFK system. It would seem pretty easy to just periodically check just a few inputs from the player.

Thoughts @Alf21 

What you have is good though, basically checks every single possible input the player can make.

GravityWolfNotAmused: I'd be willing to remove the rotation to keep Down.

mzegar: Let's get Alfs opinion on this, you could be correct with your approach here.

code/rounds/InProgressRound.cs
mzegar: That seems a bit odd, we check every single second if the alive players have changed and if the round is over.

GravityWolfNotAmused: I am assuming it's the IsRoundOver check, I will debug it in a bit, and see what it's returning. It's defiantly not the min player check.

GravityWolfNotAmused: I debugged it and found that if I revert to only calling OnKilled after changing to spectator inside of ToggledForcedSpectator it will check correctly. I have replaced it with OnKilled call. Now it will call with normal OnSecond

code/player/TTTPlayer.AFK.cs
Alf21: Just `Input.Down`, more reliable (tested,  ref <https://github.com/Neoxult/Super-Terry-Brothers/blob/7356ce349977096d6053260541d9194249c6af87/code/ui/menu/Menu.cs#L45-L63>) and nearly same as `Pressed`, but continously

GravityWolfNotAmused: So, no rotation either?

Alf21: Hmm... Based on the intention of the afk system. If it shall keep campers out as well, then without rotation. If it shall just check for afk, then with rotation (EyePos + API support, e.g. if a T is using a traitor item that interacts with another entity, not with the player pawn). Maybe, there is a way checking for mouse changes (that would avoid adding an API).

code/player/TTTPlayer.AFK.cs
```diff
@@ -0,0 +1,65 @@
...
+            {
+                bool shouldKick = ServerSettings.Instance.AFK.ShouldKickPlayers;
+
+                if (shouldKick)
+                {
+                    Log.Warning($"Steam ID: {client.SteamId}, Name: {client.Name} was kicked from the server for being AFK.");
+                    client.Kick();
+                }
+
+                if (!shouldKick)
```
Alf21, line 57:
```suggestion
                else
```

code/ui/generalhud/menu/Menu.Settings.Server.cs
```diff
@@ -22,62 +22,94 @@ internal void CreateServerSettings(PanelContent tabContent, ServerSettings serve
...
+            // TTTMinPlayers
+            CreateSettingsEntry(panelContent, "AFK Time", serverSettings.AFK.SecondsTillKick, "The length of time in minutes before a player is marked AFK.", (value) =>
+            {
+                serverSettings.AFK.SecondsTillKick = value;
+
+                ConsoleSystem.Run("ttt_serversettings_send", SettingFunctions.GetJSON(serverSettings, true));
+            });
+
+
+        }
```
Alf21, line 44:
```suggestion
        }

```

Alf21 (approved):

code/player/TTTPlayer.AFK.cs
GravityWolfNotAmused: Going to say this now, I am someone who doesn't use `else` very often, so please yell at me about this lol.

code/player/TTTPlayer.AFK.cs
GravityWolfNotAmused: I am sorry, this has gone over my head.

GravityWolfNotAmused: Made the change to use only down.

GravityWolfNotAmused:
The last few commits are for switching to just `Input.Down`, and making sure that the alive hud isn't active when in forced spectate.

GravityWolfNotAmused:
I am finished making changes beyond suggestions, I don't see any issues with anything including creating and deleting AliveHudInstance anymore.

code/rounds/InProgressRound.cs
GravityWolfNotAmused: Fixed my issue, and kept method for future events that could end the round.

code/player/TTTPlayer.AFK.cs
GravityWolfNotAmused: @Alf21 I believe what you are looking for is MouseDelta. I used it instead of Rotation. Good call.

code/player/TTTPlayer.AFK.cs
GravityWolfNotAmused: Removed, and replaed with a check `Input.MouseDelta != Vector3.Zero`

PR #229, opened 2021-09-25
Fix: Fixed inconsistencies, issues caused by sbox updates and mouse detection

Alf21:


code/ui/Hud.cs
```diff
@@ -86,16 +88,51 @@ public GeneralHud()
             }
         }
 
-        public class AliveHud : Panel
+        public class AliveHud
         {
-            public AliveHud()
+            public bool Enabled
```
mzegar, line 93: Could technically inherit from panel and then override the `Enabled` behaviour (I'm pretty sure)

Alf21: We don't need multiple panel instances, that would be incompatible anyways as there gonna be issues with z-index and ordering. With this change, you can mix panels with general hud panels, theoretically.

mzegar (approved):
Up to you on the one comment, you would need to define `_enabled`, but you'd still have to define the same overridden behaviour

code/ui/Hud.cs
mzegar: yeah nvm, there isn't any reason for this.

GravityWolfNotAmused:
Awaiting: #229 to handle possible merge conflicts with this PR.

PR #234, opened 2021-09-27
Do not use magic strings for events.

sanny-io:
This is a draft PR. I am proposing events are used in this manner rather than the way they are currently used. The benefit to this is that IntelliSense can see if it is a proper event at compile-time rather than run-time. The current usage is prone to typos.

One thing to note is that this proposal violates the editorconfig rule for keeping constants uppercase. It is done this way to keep event execution consistent with declaring events through attributes.

Consider `Event.Run(TTTEvent.Player.Died)` vs `Event.Run(TTTEvent.Player.DIED)`

I have determined the consistency with the event attributes outweighs the consistency with the editorconfig's constant uppercase rule, but if you'd like, I can change it to follow the editorconfig instead.

If I have approval to move forward with this, I can begin converting the rest of the events to be this way.


mzegar:
@Alf21 thoughts on this? I wonder if we should just go with the `public const string Died` and ignore the `public class DiedAttribute : EventAttribute` definition (since it's a decent amount of code to define for a minor change).

Alf21:
Agreed, I like the idea but we should avoid defining a boilerplate for every event. Instead, we can simply use `[Event(Name = Died)]`. But I really like this idea!

GravityWolfNotAmused:
I liked this idea as well, I have disliked looking for event names.

GravityWolfNotAmused:
@sanny-io The team is deciding how we wish to go about this, and PR #235 has been added to create the underlying code that we would use to convert all events in the system to the new format. Everything is pending until that PR is merged.

sanny-io:
It doesn't look like #235 is going to come through, so this should be ready now. Let me know what you all think.

@mzegar @Alf21 @GravityWolfNotAmused

code/global/RPCs.cs
```diff
@@ -144,19 +144,19 @@ public static void ClientClosePostRoundMenu()
         [ClientRpc]
         public static void ClientOnPlayerCarriableItemPickup(Entity carriable)
         {
-            Event.Run("tttreborn.player.carriableitem.pickup", carriable as ICarriableItem);
+            Event.Run(TTTEvent.Player.CarriableItem.PickUp, carriable as ICarriableItem);
```
GravityWolfNotAmused, line 147: Can you switch PickUp to an inventory event instead of a carriable item event?

code/global/RPCs.cs
```diff
@@ -144,19 +144,19 @@ public static void ClientClosePostRoundMenu()
...
         }
 
         [ClientRpc]
         public static void ClientOnPlayerCarriableItemDrop(Entity carriable)
         {
-            Event.Run("tttreborn.player.carriableitem.drop", carriable as ICarriableItem);
+            Event.Run(TTTEvent.Player.CarriableItem.Drop, carriable as ICarriableItem);
```
GravityWolfNotAmused, line 153: Can you switch Drop to an inventory event instead of a carriable item event?

code/global/TTTEvent.cs
```diff
@@ -0,0 +1,49 @@
...
+    public static class TTTEvent
+    {
+        public static class Player
+        {
+            public const string Died = "tttreborn.player.died";
+            public const string InitialSpawn = "tttreborn.player.initialspawn";
+            public const string Spawned = "tttreborn.player.spawned";
+            public const string TakeDamage = "tttreborn.player.takedamage";
+
+            public static class CarriableItem
```
GravityWolfNotAmused, line 14: Remove after moving events to Inventory

code/global/TTTEvent.cs
```diff
@@ -0,0 +1,49 @@
...
+
+            public static class Spectating
+            {
+                public const string Change = "tttreborn.player.spectating.change";
+            }
+        }
+
+        public static class Settings
+        {
+            public static class Instance
```
GravityWolfNotAmused, line 38: Feels like an extra step for no reason. The others devs may disagree.

code/global/TTTEvent.cs
```diff
@@ -0,0 +1,49 @@
...
+            {
+                public const string Change = "tttreborn.settings.instance.change";
```
GravityWolfNotAmused, line 40: Add this to Settings instead of Instance.

code/global/RPCs.cs
GravityWolfNotAmused: We spoke about this away from GitHub and felt that they fit best in inventory events.

GravityWolfNotAmused (changes requested):
Looks pretty good, if they are okay with all the changes I have asked for, all I ask is that you look at my branch and grab the comments for the events, or make some of your own: https://github.com/GravityWolfNotAmused/tttreborn/tree/event-changes-v2

At some point, someone will be using these events and wonder what needs to get passed, or filled in as parameters. Good to have that reminder IMO.

code/global/TTTEvent.cs
```diff
@@ -0,0 +1,49 @@
+using Sandbox;
+
+namespace TTTReborn.Globals
```
GravityWolfNotAmused, line 3: Should namespace be Globals, or Events?

sanny-io:
I can add the comments and make the rest of these changes. Although I am kind of confused. Is your branch going to be merged as well? From what was said in this PR, I thought the attributes were believed to be too much boilerplate, and I feel like there's a lot of overlap between what we're working on.

GravityWolfNotAmused:
No, I got really bored, and don't have much exposure with Attributes, so I was just fiddling around since I had #235

code/global/RPCs.cs
mzegar: Yup, I think `inventory` makes sense here.

code/global/TTTEvent.cs
mzegar: Yeah probably can get rid of instance as well.

mzegar:
Just two minor comments, thanks for the changes it'll be way nicer to reference events.

sanny-io:
I have made the requested changes.

mzegar (approved):
Thanks for the PR

GravityWolfNotAmused (approved):
LGTM, Thank you for the PR! <3

Fires1:
My only quirk with this is the actual definition location for the events, I really don't think it's necessary to have entire folder for definitions. What about extending them in their relevant areas? Such as placing the player related functions in a file near all the TTTPlayer stuff, such as `TTTPlayer.Events`. Technically would break namespace rules, but I think it's cleaner to have them in the places they are relevant in rather than defining them separately.

Alf21 (approved):
Great idea, nice changes!

Alf21:
> My only quirk with this is the actual definition location for the events, I really don't think it's necessary to have entire folder for definitions. What about extending them in their relevant areas? Such as placing the player related functions in a file near all the TTTPlayer stuff, such as `TTTPlayer.Events`. Technically would break namespace rules, but I think it's cleaner to have them in the places they are relevant in rather than defining them separately.

Thought the same, but now you know where you have to go. And you don't have namespace inconcistencies :D

mzegar:
I agree with just having them in separate files tbh. We can chat about this more if needed, but for now it's a good change.

PR #235, opened 2021-09-28
Event changes

GravityWolfNotAmused:
**Issue:**
Create a way to store, and use event types in another way other than using raw strings, so IntelliSense can pick them up for easier usage.

I had added all of the event types to TTTEvents; however, I have not changed any of the events over since #234 is open.


GravityWolfNotAmused:
Sadly this won't work: https://docs.microsoft.com/en-us/dotnet/visual-basic/misc/bc30045

I believe our best option has to be the one listed in #234 with the attribute classes.

GravityWolfNotAmused:
Tested #234 idea: https://github.com/GravityWolfNotAmused/tttreborn/tree/event-changes-v2

PR #236, opened 2021-09-29
Fix: Fixed FileSelection and replaced deprecated network classes

Alf21:


mzegar (approved):

Fires1:
My only quirk with this is the actual definition location for the events, I really don't think it's necessary to have entire folder for definitions. What about extending them in their relevant areas? Such as placing the player related functions in a file near all the TTTPlayer stuff, such as `TTTPlayer.Events`. Technically would break namespace rules, but I think it's cleaner to have them in the places they are relevant in rather than defining them separately.

Alf21:
> My only quirk with this is the actual definition location for the events, I really don't think it's necessary to have entire folder for definitions. What about extending them in their relevant areas? Such as placing the player related functions in a file near all the TTTPlayer stuff, such as `TTTPlayer.Events`. Technically would break namespace rules, but I think it's cleaner to have them in the places they are relevant in rather than defining them separately.

Uhm.. Is this connected with this PR?

PR #237, opened 2021-09-30
Ignore s&box build files.

sanny-io:
This allows developers to directly clone this project into their s&box addons folder without having to worry about committing build files from the game.


Fires1:
I like this. I hate when it tries to commit these.

Alf21 (approved):

PR #239, opened 2021-10-01
Added a "NonSpawnable" Attribute

siasur:
This adds an Attribute that should be used to prevent a TTTWeapon descendant from being spawned by a TTTWeaponRandom.

(Also I added the `Hammer.Skip` Attribute to the abstract class `TTTWeapon`.)

### DIFF
--- a/code/global/Utils.cs
+++ b/code/global/Utils.cs
@@ -142,6 +142,17 @@ namespace TTTReborn.Globals
         }
 
         /// <summary>
+        /// Loops through every type derived from the given type and collects non-abstract types that matches the given predicate.
+        /// </summary>
+        /// <param name="predicate">a filter function to limit the scope</param>
+        /// <returns>List of all available and matching types of the given type</returns>
+        public static List<Type> GetTypes<T>(Func<Type, bool> predicate)
+        {
+            return Library.GetAll<T>().Where(t => { return !t.IsAbstract && !t.ContainsGenericParameters; } )
+                .Where(predicate).ToList();
+        }
+
+        /// <summary>
         /// Get a derived `Type` of the given type by it's name (`Sandbox.LibraryAttribute`).
         /// </summary>
         /// <param name="name">The name of the `Sandbox.LibraryAttribute`</param>
--- a/code/items/NonSpawnableAttribute.cs
+++ b/code/items/NonSpawnableAttribute.cs
@@ -1,0 +1,13 @@
+using System;
+
+namespace TTTReborn.Items
+{
+
+    /// <summary>
+    /// Prevents this Entity to be spawned from a TTTWeaponRandom
+    /// </summary>
+    [AttributeUsage(AttributeTargets.Class, AllowMultiple = false, Inherited = true)]
+    public class NonSpawnableAttribute : Attribute
+    {
+    }
+}
--- a/code/items/weapons/NewtonLauncher.cs
+++ b/code/items/weapons/NewtonLauncher.cs
@@ -1,7 +1,5 @@
 using System;
-
 using Hammer;
-
 using Sandbox;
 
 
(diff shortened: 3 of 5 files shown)


GravityWolfNotAmused:
I am just pondering, but in TraitorRole, and TTTAmmoRandom there is a call to GetTypes that would require a predicate would it not?

```csharp

        /// <summary>
        /// Loops through every type derived from the given type and collects non-abstract types.
        /// </summary>
        /// <returns>List of all available types of the given type</returns>
        public static List<Type> GetTypes<T>(Func<Type, bool> predicate = null)
        {
            List<Type> types = Library.GetAll<T>().Where(t => !t.IsAbstract && !t.ContainsGenericParameters).ToList();

            if(predicate != null)
            {
                return types.Where(predicate).ToList();
            }

            return types;
        }
```

siasur:
Ja, das sollte auch zu `AmmoRandom` hinzugefügt werden. Das hatte ich übersehen beim implementieren, da ich auf `WeaponRandom` fixiert war. In TraitorRole seh ich keinen Grund auch ein predicate zu benutzen.

Please not that I did not change the original `Utils.GetTypes<T>()` but added a second one with the `predicate` parameter. Although I do not like the duplicated code that results from that, so the proposed version from you might be a good change.
But before I make any more changes I'm waiting for reaction from a maintainer. Nothing would be more annoying than to invest more time now and then find out that this change does not go in the direction the maintainers want.

If I get the green from them I'll extend the PR to include TTTAmmoRandom and remove the duplicated code that came with the parameter overloading from `Utils.GetTypes<T>()`

GravityWolfNotAmused:
Ah, then maybe this would be the same as what you are doing, but less rewriting:

```csharp

        /// <summary>
        /// Loops through every type derived from the given type and collects non-abstract types.
        /// </summary>
        /// <returns>List of all available types of the given type</returns>
        public static List<Type> GetTypes<T>()
        {
            List<Type> types = new();

            Library.GetAll<T>().ToList().ForEach(t =>
            {
                if (!t.IsAbstract && !t.ContainsGenericParameters)
                {
                    types.Add(t);
                }
            });

            return types;
        }

        /// <summary>
         /// Loops through every type derived from the given type and collects non-abstract types that matches the given predicate.
         /// </summary>
         /// <param name="predicate">a filter function to limit the scope</param>
         /// <returns>List of all available and matching types of the given type</returns>
        public static List<Type> GetTypes<T>(Func<Type, bool> predicate)
        {
            return GetTypes<T>().Where(predicate).ToList();
        }
```

siasur:
Yes, I also had something like that in my mind, although I don't really like the double `.ToList()`. My idea was more like:

```csharp
/// <summary>
        /// Loops through every type derived from the given type and collects non-abstract types.
        /// </summary>
        /// <returns>List of all available types of the given type</returns>
        public static List<Type> GetTypes<T>()
        {
            return GetTypes<T>(null);
        }

        /// <summary>
        /// Loops through every type derived from the given type and collects non-abstract types that matches the given predicate.
        /// </summary>
        /// <param name="predicate">a filter function to limit the scope</param>
        /// <returns>List of all available and matching types of the given type</returns>
        public static List<Type> GetTypes<T>(Func<Type, bool> predicate)
        {
            IEnumerable<Type> types = Library.GetAll<T>().Where(t => { return !t.IsAbstract && !t.ContainsGenericParameters; });

            if (predicate != null)
            {
                types = types.Where(predicate);
            }

            return types.ToList();
        }
```

siasur:
I was bored and decided to already implement the changes that were talked about.

Alf21 (approved):
Nice changes! If this is tested and working, we can merge this. Some time ago, we used own weapon attributes, then sbox broke because of networking and crash issues connected with these attributes. If it's working, we could get weapon attributes included again, to avoid creating weapons just to get some data for the shop.

mzegar:
> Nice changes! If this is tested and working, we can merge this. Some time ago, we used own weapon attributes, then sbox broke because of networking and crash issues connected with these attributes. If it's working, we could get weapon attributes included again, to avoid creating weapons just to get some data for the shop.

Would it be worth to bring this over into https://github.com/TTTReborn/tttreborn/pull/241 ?

Alf21:
> > Nice changes! If this is tested and working, we can merge this. Some time ago, we used own weapon attributes, then sbox broke because of networking and crash issues connected with these attributes. If it's working, we could get weapon attributes included again, to avoid creating weapons just to get some data for the shop.
> 
> Would it be worth to bring this over into #241 ?

Maybe, but then that PR definitely gonna get too big ^^

PR #240, opened 2021-10-02
Settings: Added partial settings syncing

Alf21:
This will fix a bug that gonna appear if the settings json string is getting too big.


mzegar (approved):
LGTM

PR #243, opened 2021-10-03
Scoreboard: Fixed scoreboard and removed obsolete functions

Alf21:
Finally, we got rid of PlayerScore

### DIFF
--- a/code/gamemode/Game.cs
+++ b/code/gamemode/Game.cs
@@ -3,6 +3,7 @@ using System.Linq;
 
 using Sandbox;
 
+using TTTReborn.Events;
 using TTTReborn.Globalization;
 using TTTReborn.Globals;
 using TTTReborn.Map;
@@ -104,7 +105,7 @@ namespace TTTReborn.Gamemode
 
             TTTPlayer player = new();
             client.Pawn = player;
-            player.InitialRespawn();
+            player.InitialSpawn();
 
             base.ClientJoined(client);
         }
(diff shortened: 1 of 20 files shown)


Alf21:
PR has to wait for <https://github.com/Facepunch/sbox-issues/issues/1032>

code/gamemode/Game.cs
```diff
@@ -115,9 +116,19 @@ public override void ClientDisconnect(Client client, NetworkDisconnectionReason
...
 
+            Event.Run(TTTEvent.Player.Disconnected, client.SteamId, reason);
+
+            ClientClientDisconnect(client.SteamId, reason);
+
             base.ClientDisconnect(client, reason);
         }
 
+        [ClientRpc]
+        public static void ClientClientDisconnect(ulong steamId, NetworkDisconnectionReason reason)
```
mzegar, line 127: `ClientClient` here

code/player/TTTPlayer.Commands.Server.cs
```diff
@@ -22,7 +22,7 @@ public static void RespawnPlayer()
 
             TTTPlayer player = ConsoleSystem.Caller.Pawn as TTTPlayer;
 
-            if (!player.IsValid() || ConsoleSystem.Caller.GetScore<bool>("forcedspectator", false))
+            if (!player.IsValid() || ConsoleSystem.Caller.GetValue<bool>("forcedspectator", false))
```
mzegar, line 25: I wonder if we should hardcore these strings somewhere. `forcedspectator`

mzegar (commented):
I just did a quick review on my phone. You could also use `UserId` instead of `SteamId` but both are unique anyways.

code/gamemode/Game.cs
Alf21: Yea, sadly, Intended :D

code/player/TTTPlayer.Commands.Server.cs
Alf21: Like in the event system? Could do so if we plan introducing more of these, currently, "forcedspectator" is the only one.

Alf21:
> I just did a quick review on my phone. You could also use `UserId` instead of `SteamId` but both are unique anyways.

UserId and NetworkIdent are reused Ids, steamid working as intended (if a user reconnects and the list didn't got cleaned up, that's no problem with steamid)

mzegar:
> > I just did a quick review on my phone. You could also use `UserId` instead of `SteamId` but both are unique anyways.
> 
> UserId and NetworkIdent are reused Ids, steamid working as intended (if a user reconnects and the list didn't got cleaned up, that's no problem with steamid)

`User ID of this client. These IDs will never be reused, even for the same player after rejoin.`

mzegar (approved):
LGTM

mzegar:
Merging since game mode is completely busted without this. Thanks.

Issue #192, opened 2021-08-15
`var == null` vs `var is null`


Alf21:
I saw you used `is null` earlier, we should probably just stick to one version unless there is some underlying difference. (which doesn't seem to be the case based on a quick google search)

I think we mostly do `== null` or `!= null` is our codebase, but I could be wrong.

_Originally posted by @mzegar in https://github.com/TTTReborn/tttreborn/pull/189#discussion_r689144058_

mzegar:
I had the same idea haha, I'll close the other issue https://github.com/TTTReborn/tttreborn/issues/191

Fires1:
`is` should be reserved for type checking `is Entity`...`is TTTxxxx`

Null checks should always be `== null` or with null coalescing.

mzegar:
If someone wants to take this on as a first issue, we should use `== null` whenever checking if something is null. We should use `is` for type checking.

PR #246, opened 2021-10-04
Replaced "is null" Type Checking with "== null" (#192)

TheClonker:
Hey, wanted to get my feet wet with s&box and C# after some experience with GMod (TTT) Addons.

This should cover all `is null` checks currently in the Code as described in Issue #192. For reference i just checked against the following RegExp `/is\s+(?:not\s+)?null/`.

I think it would be a good idea to enforce this in the CI, but i don't know enough about dotnet-format if it could be enforced with that alone. And I didn't wanna tinker with the config without your input 😃

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -63,7 +63,7 @@ dotnet_style_prefer_conditional_expression_over_assignment = true:silent
 dotnet_style_prefer_conditional_expression_over_return = true:silent
 dotnet_style_prefer_inferred_anonymous_type_member_names = true:suggestion
 dotnet_style_prefer_inferred_tuple_names = true:suggestion
-dotnet_style_prefer_is_null_check_over_reference_equality_method = false:warning
+dotnet_style_prefer_is_null_check_over_reference_equality_method = true
 dotnet_style_prefer_simplified_boolean_expressions = true:suggestion
 dotnet_style_prefer_simplified_interpolation = true:suggestion
 
--- a/.editorconfig
+++ b/.editorconfig
@@ -63,7 +63,7 @@ dotnet_style_prefer_conditional_expression_over_assignment = true:silent
 dotnet_style_prefer_conditional_expression_over_return = true:silent
 dotnet_style_prefer_inferred_anonymous_type_member_names = true:suggestion
 dotnet_style_prefer_inferred_tuple_names = true:suggestion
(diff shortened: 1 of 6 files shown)


Alf21 (approved):
Nice, thanks! Regarding CI / Linter / editconfigs, maybe @Histalek can provide further details / implement this.

Histalek:
https://github.com/TTTReborn/tttreborn/blob/e41601c832dc6d4eb85975dfd46c387875e54ba5/.editorconfig#L66 should be the line that needs changing. To enforce this in CI it should be enough to set it to `false:warning`.

TheClonker:
Hey, thanks for the quick response 🙂 @Histalek @Alf21

I updated the .editorconfig line but wasn't able to verify that `xyz is null` was marked as warning. 
But that could be an error on my side because i didn't setup Visual Studio correctly yet. Maybe one of you can verify that on your side ?

mzegar:
https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/style-rules/ide0041

I don't think there is an editor config for what we want.

.editorconfig
```diff
@@ -63,7 +63,7 @@ dotnet_style_prefer_conditional_expression_over_assignment = true:silent
 dotnet_style_prefer_conditional_expression_over_return = true:silent
 dotnet_style_prefer_inferred_anonymous_type_member_names = true:suggestion
 dotnet_style_prefer_inferred_tuple_names = true:suggestion
-dotnet_style_prefer_is_null_check_over_reference_equality_method = true:suggestion
+dotnet_style_prefer_is_null_check_over_reference_equality_method = false:warning
```
mzegar, line 66: Change this to `true` and should be good to merge.

Histalek:
> [docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/style-rules/ide0041](https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/style-rules/ide0041)
> 
> I don't think there is an editor config for what we want. We would want to change the value to `true` anyways.

whoops you are right. It is the only config that mentions 'is null' so i simply assumed it will do what we want here

.editorconfig
TheClonker: Done

mzegar (approved):

PR #247, opened 2021-10-04
Fix: `Client.GetInt/Value` early init

Alf21:


mzegar:
Will this be needed after https://github.com/Facepunch/sbox-issues/issues/1044 is fixed?

code/ui/generalhud/scoreboard/Scoreboard.cs
```diff
@@ -68,11 +68,6 @@ public Scoreboard() : base()
                 AddScoreboardGroup(defaultScoreboardGroup.ToString());
             }
 
-            foreach (Client client in Client.All)
-            {
-                AddClient(client);
-            }
```
mzegar, line 74: We want to keep this for when the UI hot reloads.

PR #262, opened 2021-10-17
Add nice and readable Item names to the UI

siasur:
This PR adds nice and readable names for Weapons, Equipment and Perks.
 - [x] Inspecting a corpse won't show the internal entity name (with the appended number) anymore
 - [x] Inventory doesn't show the LibraryName anymore
 - [x] Shop and Shopeditor show a nice name instead of the LibraryName #

### DIFF
--- a/code/items/ShopItemData.cs
+++ b/code/items/ShopItemData.cs
@@ -2,6 +2,7 @@ using System;
 
 using Sandbox;
 
+using TTTReborn.Extensions;
 using TTTReborn.Player;
 
 namespace TTTReborn.Items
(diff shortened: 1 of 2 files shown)


mzegar:
> I really like the idea, but sadly, I don't think that this is the way to go as well. I'd like using the already implemented multilingual (globalization) system. There, I suggest replacing the labels with translation labels and adding the weapon's attribute name into the language file. Maybe we should think in that way about a possibility to extend the already existing language files regarding upcoming add-ons support (maybe.........)

Good point, completely forgot about that.

PR #279, opened 2021-10-24
FEA: Allow hint panel to update on each tick

mzegar:


mzegar:
Brought in some of your changes over from https://github.com/TTTReborn/tttreborn/pull/259 , specifically the `TTTPlayer.Hints.cs `. I got Fires to look at both and he thinks this interface is a bit easier to use. If we run into crazy performance problems feel free to replace this, but I genuinely think this will be the same performance and the interface is a bit easier to interact with.

code/player/PlayerCorpse.cs
```diff
@@ -137,11 +138,13 @@ public ConfirmationData GetConfirmationData()
 
         public float HintDistance => 80f;
 
+        public string TextOnTick => TTTLanguage.ActiveLanguage.GetFormattedTranslation(IsIdentified ? "CORPSE_INSPECT" : "CORPSE_IDENTIFY", new object[] { Input.GetKeyWithBinding("+iv_use").ToUpper() });
```
Alf21, line 141: Does this even update in the Tick method? If not, then this won't work and we should use the other approach. If it does work, then the corresponding function is executed every tick which means there will be a lookup every tick for every hint which I don't wanna like, not just performance-based, but also coding-wise as this does not offer the possibility to alternate the text on the screen, e.g. by another add-on later.

Alf21 (commented):
Tbh, I can't see the advantage taking this PR instead of the other. This PR is imo slower in performance and more static. We can remove lot of savety-checks from the other PR as well to keep things simple, but even that would offer the way a) to modify stuff from outside b) keeping internal things (like .TextOnTick) away from the TTTPlayer.Hints.cs file and would avoid to have to access the language table manually (which shouldn't be used if we have UI stuff for that as well which handles updating automatically).

mzegar:
> Tbh, I can't see the advantage taking this PR instead of the other. This PR is imo slower in performance and more static. We can remove lot of savety-checks from the other PR as well to keep things simple, but even that would offer the way a) to modify stuff from outside b) keeping internal things (like .TextOnTick) away from the TTTPlayer.Hints.cs file and would avoid to have to access the language table manually (which shouldn't be used if we have UI stuff for that as well which handles updating automatically).

1. The performance of this PR is negligible, we are talking fetching from a O(1) constant hashmap alongside the internals of the YOGA UI system they use which will only repaint if the UI changes. There is no way to know the internals, and even if there was a difference, I'd assume it'd be impossible to even register the performance difference.
2. I think the biggest thing for me is the simplified interface, alongside not having to create UI inside the game logic part of the codebase. Everything UI-related remains in the UI section of the codebase for this changelist. Whereas, in the other PR we are modifying UI outside of the UI codebase.

Unfortunately, we have nobody else to comment on this (other than Fires), it would be cool to get more people to examine both systems and give their opinion. If you feel like your changelist is the way to go, feel free to close this and merge your PR. You have put the most amount of time into this project so far, so feel free to make the executive decision.

Alf21:
Alright, I understand your point of view and yea I agree, keeping UI and code separated should be the way to go.

Then, please check the comments / requests I made :)

Alf21:
And how about combining both PRs a bit, returning `TranslationData` instead of accessing the language data itself should keep things even more simple. Tho, as everything should be translatable, we don't have to worry about setting static text

mzegar:
Yeah I'm fine with whatever PR, but prefer this one. Made the changes you've suggested. Feel free to delete the branch on https://github.com/TTTReborn/tttreborn/pull/259

Alf21 (approved):

PR #287, opened 2021-10-28
FEA: Add hands

mzegar:
image

image

image


code/items/equipments/Hands.cs
```diff
@@ -0,0 +1,140 @@
...
+        protected override void OnDestroy()
+        {
+            DropEntity();
+
+            base.OnDestroy();
+        }
+
+        public override void SimulateAnimator(PawnAnimator anim)
+        {
+            if (!IsServer)
```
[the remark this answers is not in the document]
mzegar: Don't want to call base in this case, it sets the hold types.

code/items/equipments/Hands.cs
```diff
@@ -0,0 +1,230 @@
...
+using TTTReborn.Player;
+
+namespace TTTReborn.Items
+{
+    interface IGrabbable
+    {
+        public bool IsHolding { get; }
+        public void Grab(TTTPlayer player, TraceResult tr);
+        public void Drop();
+        public void Update(TTTPlayer player);
```
Alf21, line 14:
```suggestion
        bool IsHolding { get; }
        void Grab(TTTPlayer player, TraceResult tr);
        void Drop();
        void Update(TTTPlayer player);
```

code/items/equipments/Hands.cs
```diff
@@ -0,0 +1,230 @@
...
+        }
+
+        public void Update(TTTPlayer player) { }
+    }
+
+    class GrabbableRagdoll : IGrabbable
+    {
+        private PhysicsBody _handPhysicsBody;
+        private WeldJoint _holdJoint;
+
```
Alf21, line 52:
```suggestion
```

code/items/equipments/Hands.cs
```diff
@@ -0,0 +1,230 @@
...
+        public GrabbableRagdoll()
+        {
+            _handPhysicsBody = PhysicsWorld.AddBody();
+            _handPhysicsBody.BodyType = PhysicsBodyType.Keyframed;
+        }
+
+        public void Grab(TTTPlayer player, TraceResult tr)
+        {
+            Transform attachment = player.GetAttachment("middle_of_both_hands")!.Value;
```
Alf21, line 67: > "middle_of_both_hands"

We should make this a const :)

code/items/equipments/Hands.cs
```diff
@@ -0,0 +1,230 @@
...
+        public override string ViewModelPath => "";
+        public override bool CanDrop() => false;
+
+        private const float MAX_PICKUP_MASS = 205;
+        private const float MAX_PICKUP_DISTANCE = 75;
+        private Vector3 MAX_PICKUP_SIZE = new(50, 50, 50);
+
+        private IGrabbable GrabbedEntity;
+        private bool IsHoldingEntity => GrabbedEntity != null && GrabbedEntity.IsHolding;
+
```
Alf21, line 109:
```suggestion
```

code/items/equipments/Hands.cs
```diff
@@ -0,0 +1,230 @@
...
+                    TryGrabEntity(player);
+                }
+                else if (Input.Released(InputButton.Attack2))
+                {
+                    GrabbedEntity?.Drop();
+                }
+                else if (IsHoldingEntity)
+                {
+                    GrabbedEntity?.Update(player);
+                }
```
Alf21, line 145: I'd remove the `else `, otherwise the `GrabbedEntity` isn't updated in the frame the player pressed `InputButton.Attack1` but already holds an item. Doing this, you should return after both previous else-if checks.

code/items/equipments/Hands.cs
```diff
@@ -0,0 +1,230 @@
...
+            {
+                return;
+            }
+
+            // Ignore any size/mass restrictions when picking up a corpse.
+            if (tr.Entity is PlayerCorpse)
+            {
+                GrabbedEntity = new GrabbableRagdoll();
+                GrabbedEntity.Grab(player, tr);
+                return;
```
Alf21, line 183:
```suggestion

                return;
```

Alf21 (changes requested):
Do you set `GrabbedEntity` somewhere to null and previously destroying it?

mzegar:
@Alf21 this should be ready.

1. How do we want to handle a body that is attached to a wall via rope, then you pickup the body and start moving away far. Probably need to check the distance from the attachment point to the person grabbing the body?
2. I'm going to look into throwing props, probably just a light throw for now.
3. Did we want a different button for the secondary action for when you are holding an entity?

mzegar:
kk polish is finished, should be really GTG.

code/items/equipments/Hands.GrabbableProp.cs
```diff
@@ -0,0 +1,76 @@
...
+            GrabbedEntity.SetParent(null);
+            GrabbedEntity.EnableHideInFirstPerson = true;
+            GrabbedEntity.Velocity += _owner.EyeRot.Forward * THROW_FORCE;
+
+            _ = WaitForAnimationFinish();
+        }
+
+        private async Task WaitForAnimationFinish()
+        {
+            await GameTask.DelaySeconds(0.6f);
```
Alf21, line 72: Please add a fail-safe like we are doing in other GameTasks (see latest PR)

code/items/equipments/Hands.GrabbableProp.cs
```diff
@@ -0,0 +1,76 @@
...
+            GrabbedEntity = null;
+        }
+
+        public void Update(TTTPlayer player)
+        {
+            // If the entity is destroyed drop it.
+            if (!GrabbedEntity?.IsValid ?? true)
+            {
+                Drop();
+                return;
```
Alf21, line 48:
```suggestion

                return;
```

code/items/equipments/Hands.GrabbableProp.cs
```diff
@@ -0,0 +1,76 @@
...
+            }
+
+            // If the entity gets another owner (i.e weapon pickup) drop it.
+            if (GrabbedEntity?.Owner != null)
+            {
+                Drop();
```
Alf21, line 54:
```suggestion

                Drop();
```

code/items/equipments/Hands.GrabbableProp.cs
```diff
@@ -0,0 +1,76 @@
+using System;
+using System.Threading.Tasks;
+
+using Sandbox;
+
+using TTTReborn.Player;
+
+namespace TTTReborn.Items
+{
+
```
Alf21, line 10:
```suggestion
```

code/items/equipments/Hands.cs
```diff
@@ -0,0 +1,191 @@
...
+                    }
+                    else
+                    {
+                        PushPlayer(player);
+                    }
+                }
+                else if (Input.Pressed(InputButton.Reload) && IsHoldingEntity)
+                {
+                    GrabbedEntity?.SecondaryAction();
+                }
```
Alf21, line 77: I think it would be more intuitive using the grap action to throw if you're holding an entity, so we replace `InputButton.Reload` with `InputButton.Attack1` moving it with a check into your first scope

code/items/equipments/Hands.cs
```diff
@@ -0,0 +1,191 @@
...
+
+                GrabbedEntity?.Update(player);
+            }
+        }
+
+        private void PushPlayer(TTTPlayer player)
```
Alf21, line 83: I'd change it into `PushEntity` with the option to push entities (map entities) as well.

code/items/equipments/Hands.cs
```diff
@@ -0,0 +1,191 @@
...
+            player.SetAnimInt("holdtype_handedness", 0);
+
+            tr.Entity.Velocity += player.EyeRot.Forward * PUSHING_FORCE;
+
+            _ = WaitForAnimationFinish();
+        }
+
+        private async Task WaitForAnimationFinish()
+        {
+            await GameTask.DelaySeconds(0.6f);
```
Alf21, line 107: Fail-Safe check

code/items/equipments/Hands.cs
```diff
@@ -0,0 +1,191 @@
...
+            // Cannot pickup items held by other players.
+            if (tr.Entity.Parent != null)
+            {
+                return;
+            }
+
+            switch (tr.Entity)
+            {
+                case PlayerCorpse:
+                    GrabbedEntity = new GrabbableCorpse(player, tr.Entity as PlayerCorpse, tr.Body, tr.Bone);
```
Alf21, line 143: You're switching over `tr.Entity`, here checking for `PlayerCorpse`, but then casting it into a `PlayerCorpse`?

code/items/equipments/Hands.cs
```diff
@@ -0,0 +1,191 @@
...
+                    break;
+                case ModelEntity model:
+                    if (!model.CollisionBounds.Size.HasGreatorOrEqualAxis(MAX_PICKUP_SIZE) && model.PhysicsGroup.Mass < MAX_PICKUP_MASS)
```
Alf21, line 146: Maybe as a TODO, not needed to be done here: Add an animation that the player tried to pickup something.

Alf21 (changes requested):
Two additional things:
1. Please remove the magnetostick as well :)
2. Please add `particles/rope.vpcf` to the precache

code/items/equipments/Hands.cs
mzegar: Sure, I'll add an issue on this when merged.

Alf21 (approved):
Nice, great stuff 👍

PR #300, opened 2021-10-29
Meta: Removed Test1 perk

Alf21:
This perk was created to test stackable perks. As we have Bodyarmor and Radar now, this test item isn't needed anymore.


mzegar:
LGTM

mzegar (approved):

PR #304, opened 2021-10-30
Fixes: Fixed UI flickering bug, C4 randomness, precache and autoclose bug, added spawning ammo and tried to fix edge-case in voice chat

Alf21:


mzegar (approved):
LGTM

PR #305, opened 2021-10-30
Flashlight: Improved Flashlight rendering and shadow issue

Alf21:


mzegar (approved):
LGTM

Alf21:
This will disable the player's shadow in first person! It isn't the optimal solution as you can leak your position without knowing it when hiding, but as there is currently no other way I'm aware of, we have to deal with it.

PR #308, opened 2021-10-31
Fixes: C4, flashlight tweakings, and precache and implemented #256

Alf21:


code/gamemode/Game.cs
```diff
@@ -26,13 +26,17 @@ public partial class Game : Sandbox.Game
...
+        [ConVar.Replicated("ttt_debug")]
         public bool Debug { get; set; } = false;
 
         public Game()
         {
             Instance = this;
 
+            Utils.RNG.Next();
+            Utils.RNG.Next();
+            Utils.RNG.Next();
```
mzegar, line 38: What is this for?

Alf21: Warmup of the RNG, if it's not needed in C#, feel free to remove :)

mzegar: nah shouldn't be needed, never actually seen something like this before. Was it done in .lua before?

Alf21: yea :D

mzegar (approved):

PR #309, opened 2021-10-31
Map: Added `MapHandler` to reset the map

Alf21:


code/map/MapHandler.cs
```diff
@@ -0,0 +1,100 @@
...
+                {
+                    randomWeapons.Add(weaponRandom); // See above comment.
+                }
+                else if (entity is TTTLogicButton button)
+                {
+                    button.Cleanup();
+                }
+
+                entity.RemoveAllDecals();
+            }
```
mzegar, line 60: So I wonder if the `Delete` of the entity causes the `collection was modified`. It's just a guess, you'd have two options here if you want to test it.

1. Iterate through the `Entity.All` array backwards (go from end to the start)
2. Add the entities you plan to delete to a new array, then delete all at the end.

mzegar (approved):
Nvm, deleted my last comment, will be good for the Hands PR.

PR #311, opened 2021-10-31
Dont allow picking up of non dynamic entities

mzegar:


Alf21 (approved):
Guess you tested it :)

PR #323, opened 2021-11-03
Settings: Added auto-generated settings

Alf21:
We have to think about where we wanna store non persistent settings that are just saved every session like ForcedSpectator (in this PR removed).


code/ui/generalhud/menu/Menu.Settings.Generator.cs
```diff
@@ -0,0 +1,146 @@
...
+                                            }
+
+                                            break;
+                                    }
+
+                                    break;
+                                }
+                            }
+                        }
+                    }, categoryName.ToLower());
```
mzegar, line 141: We should break up each of these cases into a separate method, it's a bit of a huge method.

mzegar:
LGTM, yeah I think a lot easier to read

mzegar (approved):

Issue #321, opened 2021-11-02
Game ends up as a tie on start


mzegar:
## Prerequisites

- [X] I have searched [existing bug reports](https://github.com/TTTReborn/ttt-reborn/issues?q=label%3Aissue%2Fbug-report) and confirmed my issue is not already reported.

## Summary

https://streamable.com/vnoqyo

No repro information

PR #325, opened 2021-11-03
Fix: Refactor round logic, fix round starting bug, fix quickshop exception

mzegar:
- Going to run this through some testing
- I managed to repro a similar bug to #321 so this might fix it, if not, it fixes a different bug
- Fixes an exception in quickshop


code/rounds/InProgressRound.cs
```diff
@@ -60,9 +59,9 @@ protected override void OnStart()
             {
                 foreach (Client client in Client.All)
                 {
-                    if (client.Pawn is not TTTPlayer player)
+                    if (client is not TTTPlayer player)
                     {
-                        continue;
+                        return;
                     }
```
Alf21, line 65: This is always true

code/rounds/InProgressRound.cs
```diff
@@ -60,9 +59,9 @@ protected override void OnStart()
             {
                 foreach (Client client in Client.All)
                 {
-                    if (client.Pawn is not TTTPlayer player)
+                    if (client is not TTTPlayer player)
                     {
-                        continue;
+                        return;
```
Alf21, line 64: sure you wanna return?

code/rounds/InProgressRound.cs
```diff
@@ -175,11 +161,11 @@ private TTTTeam IsRoundOver()
         private void AssignRoles()
         {
             // TODO: There should be a neater way to handle this logic.
-            int traitorCount = (int) Math.Max(Players.Count * 0.25f, 1f);
+            int traitorCount = (int) Math.Max(_startingRoundPlayers.Count * 0.25f, 1f);
 
             for (int i = 0; i < traitorCount; i++)
             {
-                List<TTTPlayer> unassignedPlayers = Players.Where(p => p.Role is NoneRole).ToList();
+                List<TTTPlayer> unassignedPlayers = _startingRoundPlayers.Where(p => p.Role is NoneRole).ToList();
```
Alf21, line 168: make sure, these players still exists (could have disconnected meanwhile)

mzegar: Let me know if what I have now is fine.

code/rounds/InProgressRound.cs
```diff
@@ -201,6 +187,8 @@ private void AssignRoles()
                     player.SendClientRole();
                 }
             }
+
+            _areRolesAssigned = true;
```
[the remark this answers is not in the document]
mzegar: Yeah, so the death of a player can trigger this event, I think that's why #321 occured.

code/ui/alivehud/damageindicator/DamageIndicator.cs
```diff
@@ -22,10 +22,9 @@ public DamageIndicator() : base()
             Instance = this;
 
             StyleSheet.Load("/ui/alivehud/damageindicator/DamageIndicator.scss");
-
             Style.SetBackgroundImage(Texture.Load("/ui/damageindicator/default.png"));
-
             Style.ZIndex = -1;
+            Style.Dirty();
```
mzegar, line 27: I think a s&box bug to log here, I'm getting different behavior with this here.

code/rounds/InProgressRound.cs
```diff
@@ -75,11 +75,7 @@ protected override void OnStart()
                         player.SetHealth(player.MaxHealth);
                     }
 
-                    if (!player.IsForcedSpectator)
-                    {
-                        SetLoadout(player);
-                        _startingRoundPlayers.Add(player);
-                    }
+                    SetLoadout(player);
```
Alf21, line 78: I'd still give the loadout on spawn regarding death and the possibility searching other weapons which making it more obvious and leading to more weapons in game

code/rounds/InProgressRound.cs
```diff
@@ -160,12 +142,13 @@ private TTTTeam IsRoundOver()
...
-            // TODO: There should be a neater way to handle this logic.
-            int traitorCount = (int) Math.Max(_startingRoundPlayers.Count * 0.25f, 1f);
+            List<TTTPlayer> alivePlayers = Utils.GetAlivePlayers();
+
+            int traitorCount = (int) Math.Max(alivePlayers.Count * 0.25f, 1f);
 
             for (int i = 0; i < traitorCount; i++)
             {
-                List<TTTPlayer> unassignedPlayers = _startingRoundPlayers.Where(p => p.Role is NoneRole).ToList();
+                List<TTTPlayer> unassignedPlayers = alivePlayers.Where(p => p?.Role is NoneRole).ToList();
```
Alf21, line 151: Did you tested this? Are forced spectators alive? And why `p?`

code/ui/generalhud/inspectmenu/InspectMenu.cs
```diff
@@ -211,7 +211,13 @@ private void UpdateCurrentInspectDescription()
...
+            if (!_playerCorpse.IsValid())
+            {
+                Enabled = false;
+                return;
+            }
+
+            if (Enabled && _playerCorpse.Transform.Position.Distance(Local.Pawn.Owner.Position) > 100f)
             {
                 Enabled = false;
             }
```
Alf21, line 223: can't this be combined?

code/rounds/InProgressRound.cs
mzegar: Removed `?`, yeah I've been doing some intensive testing making sure this large logic change doesn't mess things up.

code/rounds/InProgressRound.cs
mzegar: Hmm not sure if I fully understand, could you rephrase or PM me?

Previously we were setting the loadout twice which I want to avoid. The `SetLoadout(...)` call also happens right after we either respawn the player or set their hp to max.

code/rounds/InProgressRound.cs
```diff
@@ -156,7 +124,7 @@ private TTTTeam IsRoundOver()
         {
             List<TTTTeam> aliveTeams = new();
 
-            foreach (TTTPlayer player in Players)
+            foreach (TTTPlayer player in Utils.GetAlivePlayers())
```
Alf21, line 127: TTT also has add-ons like spectator deathmatch, which is a mod that adds the ability to offer dead players the possibility to join a deathmatch while others are still fighting.

With your change, this won't be possible anymore as the specdm player would be alive as well. That's additionally why I'd like to get track of the current player in a round

code/rounds/InProgressRound.cs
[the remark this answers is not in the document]
Alf21: Wouldn't it be better then to assign the roles in PreparingRound before switching to the inprogress round (or setting it as the current one), to keep it clean

code/ui/generalhud/inspectmenu/InspectMenu.cs
```diff
@@ -211,9 +211,10 @@ private void UpdateCurrentInspectDescription()
 
         public override void Tick()
         {
-            if (Enabled && _playerCorpse?.Transform.Position.Distance(Local.Pawn.Owner.Position) > 100f)
+            if (!Enabled || !_playerCorpse.IsValid() || _playerCorpse.Transform.Position.Distance(Local.Pawn.Owner.Position) > 100f)
             {
                 Enabled = false;
+                return;
```
Alf21, line 217:
```suggestion

                return;
```

code/rounds/PreRound.cs
```diff
@@ -63,7 +129,14 @@ private static async Task StartRespawnTimer(TTTPlayer player)
...
         {
-            AddPlayer(player);
+            if (player.IsForcedSpectator)
+            {
+                return;
+            }
+
+            Extensions.Log.Debug($"Added Hands to {player.Client.Name}");
+
+            player.Inventory.TryAdd(new Hands(), true);
```
mzegar, line 139: @Alf21 How do we want to handle this, we end up calling `Player.Respawn()` on everyone (including forced spectators) which gives the spectators `Hands()`.

1. Inside of `Respawn()` we handle the `IsForcedSpectator` case, so I think we still want to call `Respawn()` on them
2. We probably want to be able to give items to dead players, so no internal check inside of `TryAdd()`

code/map/MapHandler.cs
```diff
@@ -29,6 +31,10 @@ public MapHandler()
                 {
                     ModelEntityDataList.Add(ModelEntityData.Create(entity as ModelEntity));
                 }
+                else if (entity is TTTWeaponRandom)
+                {
+                    RandomWeaponCount += 1;
```
Alf21, line 36:
```suggestion
                    RandomWeaponCount++;
```

code/rounds/InProgressRound.cs
```diff
@@ -15,10 +15,14 @@
...
 namespace TTTReborn.Rounds
 {
-    public class InProgressRound : BaseRound
+    public partial class InProgressRound : BaseRound
     {
         public override string RoundName => "In Progress";
-        private List<TTTLogicButton> LogicButtons;
+
+        [Net] public List<TTTPlayer> Players { get; set; }
+        [Net] public List<TTTPlayer> Spectators { get; set; }
```
Alf21, line 23:
```suggestion
        [Net]
        public List<TTTPlayer> Players { get; set; }

        [Net]
        public List<TTTPlayer> Spectators { get; set; }
```

code/rounds/PreRound.cs
Alf21: We should just add a check in the TryAdd method to never give a weapon to a non alive player.

code/rounds/InProgressRound.cs
```diff
@@ -152,13 +100,20 @@ private static void SetLoadout(TTTPlayer player)
...
+        }
+
         private TTTTeam IsRoundOver()
         {
             List<TTTTeam> aliveTeams = new();
 
             foreach (TTTPlayer player in Players)
             {
-                if (player.Team != null && !aliveTeams.Contains(player.Team))
+                if (player.LifeState == LifeState.Alive && player.Team != null && !aliveTeams.Contains(player.Team))
```
Alf21, line 116: Seems like disconnected players aren't handled?

code/player/Inventory.cs
```diff
@@ -87,7 +87,7 @@ public bool Add(IItem item, bool makeActive = false)
         /// <returns></returns>
         public bool TryAdd(IItem item, bool makeActive = false)
         {
-            if (!Add(item, makeActive))
+            if (Owner.LifeState == LifeState.Dead || !Add(item, makeActive))
```
mzegar, line 90: @Alf21 Maybe we want to check for != LifeState.Alive here? 

There are `Dying, Respawnable, Respawning`, I wonder if we should avoid trying to give during these as well.

Alf21: yea, that should be the way to go

code/rounds/PreRound.cs
mzegar: Done, mentioned you in another comment.

Alf21 (approved):

PR #326, opened 2021-11-06
Fea: Key Bindings

mzegar:
image

I'm pretty sure I didn't forget anything important.


mzegar:
I think this is fine for now, I'd suggest putting up an issue for refactoring the menu system, it feels like the `Menu` class is a god class, we should probably break up all of these screens into sub classes.

mzegar:
This needs a refactor, unable to get a scrolling list.

Alf21:
Ah sad, thought about the same thing haha, would be good to have such a menu

mzegar:
Yeah I'll keep the branch around and see if I can fix it up.

PR #329, opened 2021-11-08
Fea: s&box options FOV

mzegar:
Use the FOV in the settings menu of s&box everywhere.

### DIFF
--- a/code/player/camera/FirstPersonSpectatorCamera.cs
+++ b/code/player/camera/FirstPersonSpectatorCamera.cs
@@ -6,11 +6,6 @@ namespace TTTReborn.Player.Camera
     {
         private const float SMOOTH_SPEED = 25f;
 
-        public override void Activated()
-        {
-            base.Activated();
-        }
-
         public override void Deactivated()
         {
             base.Deactivated();
--- a/code/player/camera/FirstPersonSpectatorCamera.cs
+++ b/code/player/camera/FirstPersonSpectatorCamera.cs
@@ -9,8 +9,6 @@ namespace TTTReborn.Player.Camera
         public override void Activated()
         {
             base.Activated();
-
-            FieldOfView = 80;
         }
 
         public override void Deactivated()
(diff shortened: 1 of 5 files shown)


code/player/camera/FirstPersonSpectatorCamera.cs
```diff
@@ -9,8 +9,6 @@ public partial class FirstPersonSpectatorCamera : Sandbox.Camera, IObservationCa
         public override void Activated()
         {
             base.Activated();
-
-            FieldOfView = 80;
```
Alf21, line 13: Then, we don't need the Activated method at all.

code/player/camera/RagdollSpectateCamera.cs
```diff
@@ -2,8 +2,50 @@
...
+
+            FocusPoint = CurrentView.Position - GetViewOffset();
+        }
+
+        public override void Update()
+        {
+            if (Local.Client == null)
+            {
+                return;
+            }
```
Alf21, line 21: Is this needed as it's nowhere used

code/player/camera/RagdollSpectateCamera.cs
```diff
@@ -2,8 +2,50 @@
...
+        }
+
+        public virtual Vector3 GetViewOffset()
+        {
+            if (Local.Client == null)
+            {
+                return Vector3.Zero;
+            }
+
+            return Input.Rotation.Forward * (-130 * 1) + Vector3.Up * (20 * 1);
```
Alf21, line 48:
```suggestion
            return Input.Rotation.Forward * -130 + Vector3.Up * 20;
```

Or does these `* 1` any sense?

Alf21 (changes requested):

code/player/camera/RagdollSpectateCamera.cs
mzegar: Just copied and pasted code from the original ragdoll controller.

Alf21 (approved):
👍

PR #330, opened 2021-11-10
Fea: Map selection

mzegar:
This is good as a first pass for a map selection system.


code/rounds/MapSelectionRound.cs
```diff
@@ -0,0 +1,54 @@
...
+        }
+
+        protected override void OnTimeUp()
+        {
+            base.OnTimeUp();
+
+            IDictionary<int, int> mapIndexToVoteCount = MapSelection.GetTotalVotesPerMapIndex(Gamemode.Game.Instance.NextMapVotes);
+            if (mapIndexToVoteCount.Count == 0)
+            {
+                Global.ChangeLevel("facepunch.flatgrass");
```
siasur, line 29: It wold be awesome if this would be configurable to a custom default/fallback map.

mzegar: Yeah that was the plan 👍

mzegar:
I'm pretty sure this works, however, we need some testing since I can't test on the same machine due to the same `PlayerId`.

code/rounds/MapSelectionRound.cs
```diff
@@ -0,0 +1,49 @@
...
+
+        protected override void OnTimeUp()
+        {
+            base.OnTimeUp();
+
+            IDictionary<long, string> playerIdMapVote = Gamemode.Game.Instance?.MapSelection?.PlayerIdMapVote;
+            IDictionary<string, int> mapToVoteCount = MapSelectionHandler.GetTotalVotesPerMap(playerIdMapVote);
+            if (mapToVoteCount.Count == 0)
+            {
+                Global.ChangeLevel(ServerSettings.Instance?.Map?.DefaultMap ?? "facepunch.flatgrass");
```
mzegar, line 30: Do we want to random if there are no votes? or just throw it to the default map. @Alf21

code/ui/generalhud/mapselectionmenu/MapSelectionMenu.cs
```diff
@@ -0,0 +1,123 @@
...
+            IDictionary<string, int> mapToVoteCount = Map.MapSelectionHandler.GetTotalVotesPerMap(playerIdMapVote);
+
+            bool hasLocalClientVoted = playerIdMapVote.ContainsKey(Local.Client.PlayerId);
+
+            // Iterate over the map panels, update vote total, showcase the current map the player has selected.
+            for (int i = 0; i < _mapPanels.Count; ++i)
+            {
+                MapPanel panel = _mapPanels[i];
+
+                panel.TotalVotes.Text = mapToVoteCount.ContainsKey(panel.MapName) ? mapToVoteCount[panel.MapName] == 1 ? $"{1} vote" : $"{mapToVoteCount[panel.MapName]} votes" : "";
```
mzegar, line 76: Need to localize this, maybe just show the number of votes.

mzegar:
Tested this with another player, need to figure out why multiple votes don't show up on the same map.

mzegar:
Alright just one comment, otherwise from my testing this should work.

mzegar:
@Alf21 if nobody votes for a map, do we want to just fetch a random one? Otherwise from my testing this should be ready to merge.

code/events/TTTevent.MapSelectionHandler.cs
```diff
@@ -0,0 +1,14 @@
+namespace TTTReborn.Events
+{
+    public static partial class TTTEvent
+    {
+        public static class MapSelectionHandler
```
Alf21, line 5: I think just `Map` would keep things simple

code/lang/packs/en-US.json
```diff
@@ -39,6 +40,8 @@
     "CORPSE_INSPECT_IDENTIFIER_WEAPON": "It appears a {0} was used to kill them.",
     "CORPSE_INSPECT_IDENTIFIER_PERK": "They were carrying a {0}.",
 
+    "VOTE_NEXT_MAP": "VOTE FOR THE NEXT MAP!",
```
Alf21, line 43: I think it would make most sense using the normal writing, no uppercase. Uppercase should be handled via code imo

code/map/MapSelectionHandler.cs
```diff
@@ -0,0 +1,63 @@
...
+        private static async Task<List<string>> GetTTTMapImages(List<string> mapNames)
+        {
+            List<string> mapPanels = new();
+            for (int i = 0; i < mapNames.Count; ++i)
+            {
+                Package result = await Package.Fetch(mapNames[i], true);
+                mapPanels.Add(result.Thumb);
+            }
+            return mapPanels;
+        }
```
Alf21, line 46: Good for now, but later, I think partial loading would be cool. So stuff is rendered and usable but getting populated with the right image if it's loaded.

code/map/MapSelectionHandler.cs
```diff
@@ -0,0 +1,63 @@
...
+            foreach (string mapName in mapVotes.Values)
+            {
+                indexToVoteCount[mapName] = !indexToVoteCount.ContainsKey(mapName) ? 1 : indexToVoteCount[mapName] + 1;
+            }
+            return indexToVoteCount;
+        }
+
+        public void OnMapImagesChanged(IDictionary<string, string> oldValue, IDictionary<string, string> newValue)
+        {
+            Event.Run(Events.TTTEvent.MapSelectionHandler.MapImagesChange);
```
Alf21, line 60: Old and new values aren't used

code/rounds/MapSelectionRound.cs
```diff
@@ -0,0 +1,60 @@
...
+        {
+            base.OnTimeUp();
+
+            IDictionary<string, string> maps = Gamemode.Game.Instance.MapSelection.MapImages;
+
+            // We failed to fetch TTT maps, fall back to default map.
+            if (maps.Count == 0)
+            {
+                Global.ChangeLevel(ServerSettings.Instance.Map.DefaultMap);
+            }
```
Alf21, line 32: I think we should add a return just in case stuff continues running

code/rounds/MapSelectionRound.cs
```diff
@@ -0,0 +1,60 @@
...
+
+            IDictionary<long, string> playerIdMapVote = Gamemode.Game.Instance.MapSelection.PlayerIdMapVote;
+            IDictionary<string, int> mapToVoteCount = MapSelectionHandler.GetTotalVotesPerMap(playerIdMapVote);
+
+            // Nobody voted, so let's change to a random map.
+            if (mapToVoteCount.Count == 0)
+            {
+                Global.ChangeLevel(maps.ElementAt(Rand.Int(0, maps.Count - 1)).Key);
+            }
```
Alf21, line 41: Same as above

code/rounds/MapSelectionRound.cs
```diff
@@ -0,0 +1,60 @@
...
+                Global.ChangeLevel(ServerSettings.Instance.Map.DefaultMap);
+            }
+
+            IDictionary<long, string> playerIdMapVote = Gamemode.Game.Instance.MapSelection.PlayerIdMapVote;
+            IDictionary<string, int> mapToVoteCount = MapSelectionHandler.GetTotalVotesPerMap(playerIdMapVote);
+
+            // Nobody voted, so let's change to a random map.
+            if (mapToVoteCount.Count == 0)
+            {
+                Global.ChangeLevel(maps.ElementAt(Rand.Int(0, maps.Count - 1)).Key);
```
Alf21, line 40: I suggest using our RNG in the Utils class as well as keeping it equal with the role assigning.

code/rounds/MapSelectionRound.cs
```diff
@@ -0,0 +1,60 @@
...
+        {
+            player.MakeSpectator();
+        }
+
+        protected override void OnStart()
+        {
+            if (Host.IsServer)
+            {
+                RPCs.ClientOpenMapSelectionMenu();
+            }
```
Alf21, line 57: Isn't this called on the client? If so, the check isn't needed. Otherwise, the RPC isn't needed

code/ui/Hud.cs
```diff
@@ -104,6 +104,7 @@ public GeneralHud()
                 AddChild<PostRoundMenu>();
                 AddChild<Scoreboard>();
                 AddChild<Menu.Menu>();
+                AddChild<MapSelectionMenu>();
```
Alf21, line 107: Wrong position, shouldn't be displayed above the menu.

code/ui/generalhud/mapselectionmenu/MapSelectionMenu.cs
```diff
@@ -0,0 +1,127 @@
...
+                return;
+            }
+
+            IDictionary<long, string> playerIdMapVote = Gamemode.Game.Instance.MapSelection.PlayerIdMapVote;
+
+            IDictionary<string, int> mapToVoteCount = Map.MapSelectionHandler.GetTotalVotesPerMap(playerIdMapVote);
+
+            bool hasLocalClientVoted = playerIdMapVote.ContainsKey(Local.Client.PlayerId);
+
+            for (int i = 0; i < _mapPanels.Count; ++i)
```
Alf21, line 82: Wouldn't a foreach loop work as well?

Alf21 (changes requested):
Can you vote multiple maps? Is it stable if not?

code/map/MapSelectionHandler.cs
mzegar: Oh, thought they needed to be there as apart of the `OnChange`

code/events/TTTevent.MapSelectionHandler.cs
```diff
@@ -0,0 +1,14 @@
...
+        {
+            /// <summary>
+            /// Updates when the map images are networked.
+            /// <para>No data is passed to this event.</para>
+            /// </summary>
+            public const string MapImagesChange = "tttreborn.game.mapimagechange";
```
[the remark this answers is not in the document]
mzegar: I moved it to a game event, I don't think we need a separate class for this.

code/map/MapSelectionHandler.cs
mzegar: Sounds reasonable 👍

code/ui/generalhud/mapselectionmenu/MapSelectionMenu.cs
mzegar: good catch, yeah it was from a previous iteration where I needed the index for something.

code/events/TTTEvent.Game.cs
```diff
@@ -16,6 +16,12 @@ public static class Game
             /// <para>Event is passed the <strong><see cref="TTTReborn.Rounds.BaseRound"/></strong> instance of the new round.</para>
             /// </summary>
             public const string RoundChange = "tttreborn.game.roundchange";
+
+            /// <summary>
+            /// Updates when the map images are networked.
+            /// <para>No data is passed to this event.</para>
```
Alf21, line 22: Remove this :)

Alf21 (approved):
Great addition!!

PR #336, opened 2021-11-13
Fea: Show buy error items, scroll quickshop, fix quick shop description

mzegar:
1. Fixed showcasing a description of an item when nothing was selected
2. Allow the quickshop to scroll if there is a huge amount of items

### DIFF
--- a/code/ui/alivehud/quickshop/QuickShop.cs
+++ b/code/ui/alivehud/quickshop/QuickShop.cs
@@ -88,8 +88,6 @@ namespace TTTReborn.UI
 
             foreach (ShopItemData itemData in shop.Items)
             {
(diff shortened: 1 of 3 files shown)


Alf21:
I like most of the changes except the displaying of not buyable items as it shall be a "QuickShop" where you don't have to check that.

Instead, that should be a thing for the FullShop imo, so you can dive into that shop if you wanna see why it's not working. I think that gonna give the best overview to the users, but could be wrong...

Maybe should be discussed on the discord, in the end, I don't care as much ^^

mzegar:
I removed the changes in terms of logic, let's do that in a different PR, these fixes are pretty good that should be ingame right now.

Alf21 (approved):

PR #345, opened 2021-11-18
Fix: Add Debug to Logger

mzegar:


code/global/LoggerExtenstion.cs
```diff
@@ -0,0 +1,17 @@
+using Sandbox;
+
+namespace TTTReborn.Globals
+{
+    public static class LoggerExtenstion
```
Alf21, line 5:
```suggestion
    public static class LoggerExtension
```

Alf21 (changes requested):

mzegar:
@Alf21 I am merging since this fixes our uncompiling gamemode. Feel free to poke me if you have any other changes you wanted.

PR #346, opened 2021-11-19
Fix: Send RPC to one person

mzegar:


Alf21:
What does it fix? It definitely breaks the weapon syncing for the first person spectators

PR #343, opened 2021-11-17
Fea: Replace use system

mzegar:

PR #347, opened 2021-11-20
Refactor: Change ammo type to use enum

mzegar:
Wait till https://github.com/TTTReborn/tttreborn/pull/343 to merge


mzegar:
Do we want to keep the library name from the previous PR? Currently not used for anything.

Alf21:
Hmm just thought about the AmmoType var at all and it doesn't make a lot of sense to me... Theoretically, just the weapons need to have a connection to what ammo they need...

mzegar:
> Hmm just thought about the AmmoType var at all and it doesn't make a lot of sense to me... Theoretically, just the weapons need to have a connection to what ammo they need...

It's a bit better in the sense that one enum defines the ammo type instead of ensuring that the string matches up in the definition of both `PistolAmmo.cs` and `Pistol.cs`. But yeah, I think having some way to ensure that both definitions depend on each other would be ideal.

mzegar:
We should change this system as per Alfs suggestion, not wanting to actively work on this section at the moment though so closing this.

PR #349, opened 2021-11-21
Settings: Fixed settings not saving and wrong settings switch data

Alf21:


code/gamemode/Game.cs
```diff
@@ -27,8 +27,8 @@ public partial class Game : Sandbox.Game
 
         public MapHandler MapHandler { get; private set; }
 
-        [ConVar.Replicated("ttt_debug")]
-        public bool Debug { get; set; } = false;
+        // [ConVar.Replicated("ttt_debug")]
+        public bool Debug { get; set; } = true;
```
Alf21, line 31: Intended as it seems that even server admin can't change this var currently ingame

mzegar (approved):

PR #352, opened 2021-11-21
Fix: Rename weapon icons

mzegar:


mzegar:
@Alf21 kk, I think we should keep all icons in a single file since we prefix the type of the icon right before, it's easy to search.

i.e `weapon_shotgun`, `perk_radar`

I mean otherwise we can do a switch statement and check the `type` and go to the according folder instead.

Alf21:
if `quickshop` is replaced with `shop` or `icons` or `items`, we are good to go

Alf21 (approved):

PR #356, opened 2021-11-27
Fix: Added some more Task error catching

Alf21:


mzegar (commented):
Looks fine to me, is there any difference with the return time typing `void` vs. `Task`?

mzegar (approved):
LGTM, just one comment above.

PR #365, opened 2021-12-26
WIP: Remove "TTT" Panel

mzegar:
Based on https://github.com/TTTReborn/tttreborn/pull/364 , let's merge that first.

We don't need it, it's annoying to have two separate definitions since a lot of base UI objects are `Sandbox.UI.Panel` and we have to downcast it to `TTTReborn.UI.Panel`


mzegar:
I'm closing this, I'm going to do this once we merge the other two branches first.

PR #375, opened 2022-01-15
Fea: Implement SWB

mzegar:
Adds Deagle, FAL, Knife, L96A1, RPG, SPAS12


code/player/TTTPlayer.Confirmation.cs
```diff
@@ -55,7 +55,7 @@ private void BecomePlayerCorpseOnServer(Vector3 force, int forceBone)
                 Rotation = Rotation
             };
 
-            corpse.KillerWeapon = LastDamageWeapon?.LibraryName;
+            // corpse.KillerWeapon = LastDamageWeapon?.LibraryName;
```
Alf21, line 58: So the waepon isn't passed anymore?

code/player/Inventory.cs
```diff
@@ -37,7 +34,7 @@ public override void DeleteContents()
             RPCs.ClientClearInventory(To.Multiple(Utils.GetClients((pl) => pl.CurrentPlayer == Owner as TTTPlayer)));
 
             Perks.Clear();
-            Ammo.Clear();
+            (Owner as TTTPlayer).Ammo.Clear();
```
Alf21, line 37: Why saving ammo on playaer instead on inventory now?

code/swb_base/AttachmentBase.cs
```diff
@@ -0,0 +1,170 @@
+using System;
```
Alf21, line 1: Do we really wanna support attachments in vanilla?

code/swb_base/InventoryBase.cs
```diff
@@ -0,0 +1,71 @@
+using System;
```
Alf21, line 1: imo should be merged with ours? Or did you just wrapped the swb system so we can use multiple add-ons? Afaik, you're referencing them directly.

code/items/weapons/generics/TTTWeaponBase.cs
```diff
@@ -0,0 +1,83 @@
+using System;
```
Alf21, line 1: Is this class / file used?

Alf21 (commented):
Such a huge PR with cool stuff, but I don't know whether we need it in that size. Especially as we have to manage updates here as well. That's why I'd prefer a system where we can add and remove such add-ons instead of maintaining them.

code/player/TTTPlayer.Confirmation.cs
mzegar: Sorry, need to go back and this this, Good catch.

code/player/Inventory.cs
mzegar: Since SWB stores ammo on the player.

code/swb_base/AttachmentBase.cs
mzegar: No, I can remove it, I wonder how much other code will change though.

code/swb_base/InventoryBase.cs
mzegar: Removed `InventoryBase.cs`, not needed.

code/items/weapons/generics/TTTWeaponBase.cs
mzegar: Yeah, everything inside of `generics` is being used. So `SWB` has multiple derived classes called `SWB_Base.WeaponBase`, `SWB_Base.WeaponBaseEntity`, `SWB_Base.WeaponBaseMelee`, etc.

I needed to make a class to derive from each of these classes so we could put our `TTT` data on them (ex. `TTTWeaponBase`, `TTTWeaponBaseMelee`, etc.). 

The other option would be to do `composition`, where a new `TTTWeaponData` class is made and then placed inside of each one of these classes. This might be a bit cleaner, but the `TTTWeaponBaseGeneric` cleans it up a bit.

mzegar:
> Such a huge PR with cool stuff, but I don't know whether we need it in that size. Especially as we have to manage updates here as well. That's why I'd prefer a system where we can add and remove such add-ons instead of maintaining them.

https://github.com/timmybo5/simple-weapon-base/issues/12

Yeah, I want SWB to basically be a git module that can be installed w/o issue. I had to modify `three` total internal SWB files for this PR, so it isn't that bad. Zombie seems to be working on getting it to a state where we won't have to touch any internals.

code/swb_base/AttachmentBase.cs
mzegar: Yeah, it's a bit too intertwined with the base code, it would have to stay.

mzegar:
Spoke with @Alf21 , he is not interested in using SWB, closing.

PR #421, opened 2022-04-16
Refactoring: GameEvent Networking

Alf21:


Alf21:
It makes networking easier and faster (lower capacity), but downside of these changes are:
- Exposed parameterless constructors that should not be used (due to sbox `Library` limitations)
- No backward compatibility with probably saved binary data
- Networking has to be done in detail per class if there is additional data

Best solution would be saving via the old system (json data), networking via the new system (binary data).

As the downside of this PR is bigger than the upside, this PR is closed.

